PR #34, opened 2016-05-17
Fix acceptance tests

billyevans:
Add removing all topics from Kafka before each Scenario.
Put back kafka-check test with empty cluster.
Some minor fixes after that.


yzernik:
+1

PR #39, opened 2016-05-24
Support local import for group-parser

ecanzonieri:
This allows to specify the group parser as: --group-parser my.module


yzernik:
+1

PR #40, opened 2016-05-26
Add under_replicated sub-command for kafka-check

billyevans:
It iterates across all brokers from the cluster and query them via jolokia port about under replicated partitions number. Returns OK if 0 of them, CRITICAL otherwise.

### DIFF
--- a/docs/source/kafka_check.rst
+++ b/docs/source/kafka_check.rst
@@ -22,7 +22,7 @@ The parameters for min_isr check are:
 
 .. code-block:: bash
 
-   $ kafka-check --cluster-type=sample_type min_isr 
+   $ kafka-check --cluster-type=sample_type min_isr
    OK: All replicas in sync.
 
 In case of min isr violations:
@@ -35,3 +35,39 @@ In case of min isr violations:
     CRITICAL: 1 partition(s) have the number of replicas in sync that is lower
     than the specified min ISR.
 
+Checking under replicated partitions
+====================================
+This kafka tool provides the ability to check and report number of under replicated
+partitions for all brokers in the cluster.
+
+The :code:`under_replicated` command checks if the number of under replicated partitions
+is equal to zero. It will report the aggregated result of under replicated partitions
+of each broker if any.
+
+The parameters specific to under_replicated check are:
+
+* :code:`--first-broker-only`: If this parameter is specified, the command will
+  check for under-replicated partitions for given broker only if it's the first
+  broker in broker-list fetched from zookeeper. Otherwise, it does nothing and succeeds.
+  If :code:`--broker-id` is also set as -1 then broker-id will be computed from given
+  data-path.
+* :code:`--jolokia-port JOLOKIA_PORT`: Port to access to jolokia collector.
+
+.. code-block:: bash
+
+   $ kafka-check --cluster-type=sample_type under_replicated
+   OK: No under replicated partitions
+
+In case of not first broker in the broker list in Zookeeper:
+
+.. code-block:: bash
+
+   $ kafka-check --cluster-type=sample_type --broker-id 3 under_replicated --first-broker-only
+   OK: Provided broker is not the first in broker-list
+
+In case where some partitions are under-replicated.
+
+.. code-block:: bash
+
+   $ kafka-check --cluster-type=sample_type under_replicated
+   CRITICAL: 2 under replicated partitions
--- a/kafka_utils/kafka_check/commands/under_replicated.py
+++ b/kafka_utils/kafka_check/commands/under_replicated.py
@@ -1,0 +1,179 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from __future__ import absolute_import
+
+import argparse
+import json
+
+import requests
+
+from kafka_utils.kafka_check import status_code
+from kafka_utils.kafka_check.commands.command import get_broker_id
+from kafka_utils.kafka_check.commands.command import KafkaCheckCmd
+
+
+DEFAULT_JOLOKIA_PORT = 8778
+
+
+class UnderReplicatedCmd(KafkaCheckCmd):
+
+    def build_subparser(self, subparsers):
+        subparser = subparsers.add_parser(
+            'under_replicated',
+            description='Check under replicated partitions for all '
+                        'brokers in cluster.',
+            help='This command will sum all under replicated partitions '
+                 'for each broker if any. It will query jolokia port for '
+                 'receive this data.',
+        )
+        subparser.add_argument(
+            '--first-broker-only',
+            action='store_true',
+            help='If this parameter is specified, it will do nothing and succeed '
+                 'on not first brokers from Kafka cluster. Set --broker-id to -1 '
+                 'to read broker-id from --data-path. Default: %(default)s',
+        )
+        subparser.add_argument(
+            '--jolokia-port',
+            type=convert_to_port,
+            default=DEFAULT_JOLOKIA_PORT,
+            help='Port to access to jolokia on each broker. Default: %(default)s',
+        )
+
+        return subparser
+
+    def run_command(self):
+        """Under_replicated command, checks number of under replicated partitions for
+        all brokers in the Kafka cluster."""
+        broker_list = self.zk.get_brokers()
+
+        if self.args.first_broker_only:
+            if self.args.broker_id is None:
+                return status_code.WARNING, 'Broker id is not specified'
+
+            if not check_run_on_first_broker(broker_list, self.args.broker_id, self.args.data_path):
+                return status_code.OK, 'Provided broker is not the first in broker-list'
+
+        under_replicated = get_under_replicated(broker_list, self.args.jolokia_port)
+
+        if len(under_replicated) == 0:
+            return status_code.OK, 'No under replicated partitions'
+        else:
+            count = 0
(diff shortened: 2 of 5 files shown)


kafka_utils/kafka_check/commands/under_replicated.py
```diff
@@ -0,0 +1,129 @@
...
+            description='Check under replicated partitions for all brokers in cluster.',
+            help='This command will sum all numbers of under replicated partitions from each broker.'
+                 'It will query jolokia port for receive this data.',
+        )
+        subparser.add_argument(
+            '--first-broker-only',
+            action='store_true',
+            help='If this parameter is specified, it will do nothing and succeed on '
+                 'not first brokers from Kafka cluster. Set --broker-id to -1 to read broker-id '
+                 'from --data-path',
```
fede1024: Consider printing the default value (true). Same for the next parameter.

kafka_utils/kafka_check/commands/under_replicated.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+def check_run_on_first_broker(broker_list, args):
+    """Returns true if the first broker in broker_list the same as in args."""
+    if args.broker_id is None:
+        raise RuntimeError("Broker id is not specified")
+
+    broker_id = args.broker_id if args.broker_id != -1 else get_broker_id(args.data_path)
+
+    return broker_id == broker_list.iterkeys().next()
```
fede1024: This line relies on the fact that the broker_list returned by the zk module is sorted, but it would probably be better to sort it again, so that this function doesn't have to rely on a particular input format.

kafka_utils/kafka_check/commands/under_replicated.py
```diff
@@ -0,0 +1,129 @@
...
+    try:
+        value = int(string)
+    except ValueError:
+        raise argparse.ArgumentTypeError(error_msg)
+    if value < MIN_PORT_NUMBER or value > MAX_PORT_NUMBER:
+        raise argparse.ArgumentTypeError(error_msg)
+    return value
+
+
+def check_run_on_first_broker(broker_list, args):
```
fede1024: We should pass the broker_id from run_command and check the presence of the argument directly there.

kafka_utils/kafka_check/commands/under_replicated.py
```diff
@@ -0,0 +1,129 @@
...
+    broker_id = args.broker_id if args.broker_id != -1 else get_broker_id(args.data_path)
+
+    return broker_id == broker_list.iterkeys().next()
+
+
+def get_under_replicated_from_broker(host, port):
+    """Query one broker for info about under replicated partitions.
+
+    :param string host: hostname of broker
+    :param int port: jolokia port for query under replicated partitnions number
```
fede1024: typo in partitions (same in the next line and next function)

fede1024:
+1

docs/source/kafka_check.rst
```diff
@@ -35,3 +35,37 @@ In case of min isr violations:
     CRITICAL: 1 partition(s) have the number of replicas in sync that is lower
     than the specified min ISR.
 
+Checking under replicated partitions
+====================================
+This kafka tool provides the ability to check number of under replicated partitions for all
```
88manpreet: Would be 'ability to check' or 'ability to report' or 'ability to check and report'?

docs/source/kafka_check.rst
```diff
@@ -35,3 +35,37 @@ In case of min isr violations:
...
+brokers in the cluster.
+
+The :code:`under_replicated` command checks if the number of under replicated partitions
+is equal to zero. This command will sum all numbers of under replicated
```
88manpreet: Don't think 2nd line is necessary. Or may be, 'It will report the aggregated result of under replicated partitions of each broker if any.'

docs/source/kafka_check.rst
```diff
@@ -35,3 +35,37 @@ In case of min isr violations:
...
+partitions from each broker.
+
+The parameters for under_replicated check are:
```
88manpreet: The parameters specific to under_replicated check are:

docs/source/kafka_check.rst
```diff
@@ -35,3 +35,37 @@ In case of min isr violations:
...
+
+* :code:`--first-broker-only`: If this parameter is specified, it will do nothing and
```
88manpreet: Did not get the first line. Maybe revisit once?

docs/source/kafka_check.rst
```diff
@@ -35,3 +35,37 @@ In case of min isr violations:
...
+   OK: No under replicated partitions
+
+In case of not first broker in the broker list in Zookeeper:
+
+.. code-block:: bash
+
+   $ kafka-check --cluster-type=sample_type --broker-id 3 under_replicated --first-broker-only
+   OK: Provided broker is not the first in broker-list
+
+In case of some partitions are under replicated:
```
88manpreet: In case where some partitions are under-replicated.

kafka_utils/kafka_check/commands/under_replicated.py
```diff
@@ -0,0 +1,135 @@
...
+DEFAULT_JOLOKIA_PORT = 8778
+
+
+class UnderReplicatedCmd(KafkaCheckCmd):
+
+    def build_subparser(self, subparsers):
+        subparser = subparsers.add_parser(
+            'under_replicated',
+            description='Check under replicated partitions for all brokers in cluster.',
+            help='This command will sum all numbers of under replicated partitions from each broker.'
```
88manpreet: will sum all under replicated partitions for each broker if any.

docs/source/kafka_check.rst
```diff
@@ -35,3 +35,37 @@ In case of min isr violations:
...
+brokers in the cluster.
+
+The :code:`under_replicated` command checks if the number of under replicated partitions
+is equal to zero. This command will sum all numbers of under replicated
+partitions from each broker.
+
+The parameters for under_replicated check are:
+
+* :code:`--first-broker-only`: If this parameter is specified, it will do nothing and
+  succeed on not first broker in the list from Zookeeper. If :code:`--broker-id` is also
```
88manpreet: I think you mean, 'The command will check for under-replicated partitions for given broker only if it's the first broker in broker-list fetched from zookeeper. Otherwise, it does nothing and succeeds.

88manpreet:
lgtm, some minor comments and then +1

billyevans:
It's fixed and ready to merge. I've made git rebase.

kafka_utils/kafka_check/commands/under_replicated.py
```diff
@@ -0,0 +1,135 @@
...
+
+DEFAULT_JOLOKIA_PORT = 8778
+
+
+class UnderReplicatedCmd(KafkaCheckCmd):
+
+    def build_subparser(self, subparsers):
+        subparser = subparsers.add_parser(
+            'under_replicated',
+            description='Check under replicated partitions for all brokers in cluster.',
```
ecanzonieri: this line is very long, can you split it at 80 chars?

kafka_utils/kafka_check/commands/under_replicated.py
```diff
@@ -0,0 +1,135 @@
...
+            help='This command will sum all under replicated partitions for each broker if any.'
```
ecanzonieri: same here

kafka_utils/kafka_check/commands/under_replicated.py
```diff
@@ -0,0 +1,135 @@
...
+    def run_command(self):
+        """Under_replicated command, checks number of under replicated partitions for
+        all brokers in the Kafka cluster."""
+        broker_list = self.zk.get_brokers()
+
+        if self.args.first_broker_only:
+            if self.args.broker_id is None:
+                return status_code.WARNING, 'Broker id is not specified'
+
+            if not check_run_on_first_broker(broker_list, self.args.broker_id, self.args.data_path):
```
ecanzonieri: what's the use case for this? when would you want/need to use this condition?

kafka_utils/kafka_check/commands/under_replicated.py
```diff
@@ -0,0 +1,135 @@
...
+    JOLOKIA_PREFIX = 'jolokia'
+
+    url = 'http://{host}:{port}/{prefix}/read/{key}'.format(
+        host=host,
+        port=port,
+        prefix=JOLOKIA_PREFIX,
+        key=UNDER_REPL_KEY,
+    )
+
+    response = requests.get(url)
```
ecanzonieri: we should have a timeout here. iirc requests default timeout is something like 1 hour.

kafka_utils/kafka_check/commands/under_replicated.py
```diff
@@ -0,0 +1,135 @@
...
+
+    response.raise_for_status()
+
+    under_replicated = json.loads(response.content)['value']
+    return sum([int(val['Value']) for val in under_replicated.values()])
```
ecanzonieri: I think that the information about what partitions are under replicated is quite important. I would rather to keep a map and in the result give the information about under-replicated partitions for each broker id.

kafka_utils/kafka_check/commands/under_replicated.py
```diff
@@ -0,0 +1,135 @@
...
+
+def get_under_replicated(broker_list, port):
+    """Go across each broker in broker_list and query under replicated partitions number.
+    Then summarize them and return.
+
+    :param dictionary broker_list: dictionary with brokers information, broker_id is key
+    :param int port: jolokia port for query under replicated partitions number
+    :returns int: total number of under replicated partitions
+    """
+    return sum(
```
ecanzonieri: As I commented above, having a total count for a cluster is very limiting. You'll have to run manual commands anyway to figure out what broker is under replicated and what partitions. I'd like to have this information included in this health check

kafka_utils/kafka_check/commands/under_replicated.py
billyevans: We want run this check only on one kafka broker. First idea was run it on controller node.
But during restart controller can be changed many times. So, we decided to run it only on the first broker from broker list from Zookeeper.

kafka_utils/kafka_check/commands/under_replicated.py
billyevans: I had similar thoughts, but decided to make just substitution for the current check with total number only as result.

billyevans:
It's fixed and ready to merge.

ecanzonieri:
+1

PR #41, opened 2016-06-08
Fix bug for No available brokers to move partition to

88manpreet:
If there are no available brokers with destination broker for eligible partition (because all brokers already have a replica for the same) it should return None.


ecanzonieri:
Can you add a test for this condition?

ecanzonieri:
+1

PR #42, opened 2016-06-13
Fix two decommissioning bugs

ecanzonieri:
Change separate_groups to extend underloaded with elements with optimal count.
Fix a bug in force_broker_decommission (missing break).
Add tests to cover both edge cases.


fede1024:
+1

PR #44, opened 2016-06-16
Fix logging in kafka cluster manager

fede1024:
By default fileConfig disables all loggers that are define before it is called. Most of our loggers are defined during application start and get disabled. This change prevent fileConfig from disabling them.


ecanzonieri:
+1

PR #47, opened 2016-06-28
Use Kafka metadata request instead of Jolokia in under replicated

billyevans:
Add acceptance tests for under_replicated command
Rename min_isr tests


fede1024:
+1

PR #48, opened 2016-07-01
Improve speed of min_isr command with kafka metadata request

billyevans:


tests/kafka_check/test_min_isr.py
```diff
@@ -13,128 +13,151 @@
...
+
+
+def test_get_min_isr_empty():
+    TOPIC_CONFIG_WITHOUT_MIN_ISR = {'version': 1, 'config': {'retention.ms': '86400000'}}
+    attrs = {'get_topic_config.return_value': TOPIC_CONFIG_WITHOUT_MIN_ISR}
+    zk_mock = mock.MagicMock(**attrs)
+
+    min_isr = get_min_isr(zk_mock, REGULAR_TOPIC_NAME0)
+
+    assert zk_mock.method_calls == [mock.call.get_topic_config(REGULAR_TOPIC_NAME0)]
```
ecanzonieri: zk_mock.get_topic_config.assert_called_once_with(REGULAR_TOPIC_NAME0)

tests/kafka_check/test_min_isr.py
```diff
@@ -13,128 +13,151 @@
...
-        code, msg = cmd.run_command()
-
-        assert code == status_code.CRITICAL
-        assert msg == "2 partition(s) have the number of replicas in sync that is lower than the specified min ISR."
+    attrs = {'get_topic_config.return_value': TOPIC_CONFIG_WITH_MIN_ISR}
+    zk_mock = mock.MagicMock(**attrs)
+
+    min_isr = get_min_isr(zk_mock, REGULAR_TOPIC_NAME0)
+
+    assert zk_mock.method_calls == [mock.call.get_topic_config(REGULAR_TOPIC_NAME0)]
```
ecanzonieri: same here, prefer to use assert methods provided by mock

tests/kafka_check/test_min_isr.py
```diff
@@ -13,128 +13,151 @@
...
-                }
-            }
-        },
+from kafka.common import PartitionMetadata
+
+from kafka_utils.kafka_check.commands.min_isr import get_min_isr
+from kafka_utils.kafka_check.commands.min_isr import process_metadata_response
+
+
+REGULAR_TOPIC_NAME0 = 'topic_0'
```
ecanzonieri: why do we have these constants? What does REGULAR_* stand for?

billyevans: Yeah, probably we get get rid of it.

tests/kafka_check/test_min_isr.py
```diff
@@ -13,128 +13,148 @@
...
+
+    zk_mock.get_topic_config.assert_called_once_with('topic_0')
+    assert min_isr is None
+
+
+def test_get_min_isr():
+    TOPIC_CONFIG_WITH_MIN_ISR = {
+        'version': 1,
+        'config': {'retention.ms': '86400000'},
+        'min.insync.replicas': '3',
```
fede1024: The 'min.insync.replicas' should be in the 'config' dictionary. This means that our get_min_isr function is broken as well.

billyevans: Nice catch!

fede1024:
+1

kafka_utils/kafka_check/commands/min_isr.py
```diff
@@ -33,47 +34,63 @@ def build_subparser(self, subparsers):
...
-            return int(config[ISR_CONF_NAME])
-        else:
-            return None
-
     def run_command(self):
         """Min_isr command, checks number of actual min-isr
         for each topic-partition with configuration for that topic."""
-        topics = self.zk.get_topics()
-        not_in_sync = 0
+        broker_list = self.zk.get_brokers()
```
ecanzonieri: why do we need to get the broker list from zookeeper, while we can just use the broker_list from the ClusterConfig?

billyevans: I can change that to use just ClusterConfig, but I will need to additional function to convert broker_list from ClusterConfig to format which KafkaClient accepts. Further more now in under_replicated command I have the same usage of it, after that it will be different. I am not sure that's worth to change it.

ecanzonieri: ClusterConfig.broker_list is already in a KafkaClient compatible format. I think that all the other tools just use the broker_list from ClusterConfig instead of reading and building the broker_list from zookeeper.

billyevans: Oh, right. I see, thank you!

kafka_utils/util/metadata.py
```diff
@@ -15,7 +15,7 @@
 from kafka import KafkaClient
 
 
-def _prepare_host_list(broker_list):
+def prepare_host_list(broker_list):
```
fede1024: I don't think this function is being used anywhere in the code. Can we remove it?

billyevans: I will use this one in under_replicated partitions after merging that. Now there is copy of it.

fede1024: I see. +1

ecanzonieri: Why do you need this function in under_replicated? Can't we use broker_list from cluster_config there?

billyevans: Good idea, I will drop it.

fede1024:
+1

ecanzonieri:
+1

PR #49, opened 2016-07-04
Use error field from metadata response

billyevans:


kafka_utils/kafka_check/commands/under_replicated.py
```diff
@@ -14,15 +14,16 @@
...
-from collections import defaultdict
-
 from kafka import KafkaClient
 
 from kafka_utils.kafka_check import status_code
 from kafka_utils.kafka_check.commands.command import get_broker_id
 from kafka_utils.kafka_check.commands.command import KafkaCheckCmd
 
 
+REPLICA_NOT_AVAILABLE = 9
```
fede1024: Can we call it REPLICA_NOT_AVAILABLE_ERROR and add a line of comment?

fede1024:
+1

PR #52, opened 2016-07-14
Update main.py

neoeahit:
Currently the error message starts from 0 and counts till n, which means it does total n+1, making it to start from 1 so that it runs till n for a total of n times


ecanzonieri:
I think this change broke tests. Can you run make test on your dev and fix the error?

neoeahit:
i fixed it, it was a formatting error.

ecanzonieri:
+1

PR #54, opened 2016-07-26
Fix empty replication group error.

ecanzonieri:
This bug makes impossible to decommission a broker that failed if there is no more metadata available to identify the origin replication group.

### DIFF
--- a/kafka_utils/kafka_cluster_manager/cluster_info/cluster_topology.py
+++ b/kafka_utils/kafka_cluster_manager/cluster_info/cluster_topology.py
@@ -21,6 +21,7 @@ from collections import OrderedDict
 
 from .broker import Broker
 from .error import BrokerDecommissionError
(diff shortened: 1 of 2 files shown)


kafka_utils/kafka_cluster_manager/cluster_info/cluster_topology.py
```diff
@@ -207,11 +210,9 @@ def _decommission_brokers_in_group(self, group):
                         broker,
                         broker.partitions,
                     )
-                    failed = True
-            if failed:
-                raise BrokerDecommissionError(
-                    "Broker decommission failed."
-                )
+                    raise BrokerDecommissionError(
+                        "Broker decommission failed."
```
88manpreet: minor ',' in the end. In fact in single line.

88manpreet:
fix and +1

PR #55, opened 2016-07-29
Bump version for decommission bug due to empty replication-group found

88manpreet:
Update version of __init__.py


ecanzonieri:
+1

PR #57, opened 2016-08-11
Improve consumer manager json ouput

Mango-J:
remove non-json part from --json output.
add itest for this change.


ecanzonieri:
+1

PR #60, opened 2016-08-18
Remove force-yes and remove pinned docker-engine

ecanzonieri:
I don't think we need to pin a docker version. The build was failing because the docker version changed and the latest is not 1.11 but 1.12.


ny2ko:
+1

PR #61, opened 2016-08-22
Fix logging bug when rg is None

fede1024:
The replication group map can contain None keys when a broker is down. This should trigger a RebalanceError, but this logging line will fail first. This change will make sure that the logging line won't fail in presence of None values.


ny2ko:
+1

PR #62, opened 2016-09-02
Fix update of sibling-count bug

88manpreet:
In the test case: Before the fix the net_imbal is 3, the original imbal is 5

### DIFF
--- a/kafka_utils/kafka_cluster_manager/cluster_info/rg.py
+++ b/kafka_utils/kafka_cluster_manager/cluster_info/rg.py
@@ -243,7 +243,7 @@ class ReplicationGroup(object):
                 sibling_distance = self.update_sibling_distance(
                     sibling_distance,
                     broker_destination,
-                    victim_partition,
+                    victim_partition.topic,
                 )
             else:
                 # Brokers are balanced or could not be balanced further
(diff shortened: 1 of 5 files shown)


kafka_utils/kafka_cluster_manager/cluster_info/rg.py
```diff
@@ -243,7 +243,7 @@ def rebalance_brokers(self):
                 sibling_distance = self.update_sibling_distance(
                     sibling_distance,
                     broker_destination,
-                    victim_partition,
+                    victim_partition.topic,
```
ecanzonieri: Nice catch!

tests/kafka_cluster_manager/cluster_topology_test.py
```diff
@@ -236,6 +238,34 @@ def test_rebalance_replication_groups(self):
             ct.brokers.keys(),
         )
 
+    def test_rebalance_brokers_for_topic_partition_imbalance(self):
```
ecanzonieri: I think we should add a unit test in rg_test.py rather than in ClusterTopology.

88manpreet:
Not able to reply to your comment.
This test builds cluster-toplogy and rebalance all brokers. So I guess its better here?

88manpreet:
Not able to reply to your comment.
This test builds cluster-toplogy and rebalance all brokers. So I guess its better here?

88manpreet:
Not able to reply to your comment.
This test builds cluster-toplogy and rebalance all brokers. So I guess its better here?

tests/kafka_cluster_manager/cluster_topology_test.py
88manpreet: Not able to reply to your comment.
This test builds cluster-toplogy and rebalance all brokers. So I guess its better here?

PR #64, opened 2016-09-21
Ensure kafka-cluster-manager commands exit cleanly on empty clusters

anthonysandrin:
Some of the kafka-cluster-manager sub-commands (`rebalance`, `decommission`, `move-brokers`) fail with exit status `1` when there are no topics in the cluster. An empty cluster is not be an error so the exit status should be `0`. This is true for all sub-commands so we can make this change globally.

### DIFF
--- a/tests/kafka_cluster_manager/cmds/command_test.py
+++ b/tests/kafka_cluster_manager/cmds/command_test.py
@@ -288,6 +288,7 @@ class TestClusterManagerCmd(object):
         )
         rg_parser = mock.MagicMock()
         cmd.run_command = mock.MagicMock()
+
         cmd.run(cluster_config, rg_parser, args)
 
         assert cmd.run_command.call_count == 0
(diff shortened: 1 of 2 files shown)


tests/kafka_cluster_manager/cmds/command_test.py
```diff
@@ -272,3 +273,21 @@ def test_reduced_proposed_plan(
...
+            get_brokers=lambda: {
+                1: {'host': 'host1'},
+                2: {'host': 'host2'},
+                3: {'host': 'host3'},
+            },
+            get_assignment=lambda: {},
+        )
+        rg_parser = mock.MagicMock()
+        cmd.run_command = mock.MagicMock()
+        cmd.run(cluster_config, rg_parser, args)
```
ecanzonieri: line break before cmd.run(..)

PR #65, opened 2016-09-21
Add set_replication_factor sub-command to kafka-cluster-manager

anthonysandrin:
`usage: kafka-cluster-manager set_replication_factor [-h] --topic TOPIC replication_factor`

This command increases or decreases the replication factor of all partitions of a topic. It tries to do so in a way that spreads replicas across replication-groups and maintains the overall balance of the cluster.

### DIFF
--- a/kafka_utils/kafka_cluster_manager/cluster_info/error.py
+++ b/kafka_utils/kafka_cluster_manager/cluster_info/error.py
@@ -20,8 +20,20 @@ class InvalidBrokerIdError(KafkaToolError):
     pass
 
 
+class InvalidBrokerError(KafkaToolError):
+    """Raised when a broker doesn't exist within a replication-group."""
+    pass
+
+
 class InvalidPartitionError(KafkaToolError):
     """Raised when a partition tuple (topic, partition) doesn't exist in the cluster"""
+    pass
+
+
+class InvalidReplicationFactorError(KafkaToolError):
+    """Raised when an operation would result in an replication factor that is
+    too small or too large for the cluster.
+    """
     pass
 
 
--- a/kafka_utils/kafka_cluster_manager/cluster_info/error.py
+++ b/kafka_utils/kafka_cluster_manager/cluster_info/error.py
@@ -20,11 +20,6 @@ class InvalidBrokerIdError(KafkaToolError):
     pass
 
 
-class InvalidBrokerError(KafkaToolError):
-    """Raised when a broker doesn't exist within a replication-group."""
-    pass
-
-
 class InvalidPartitionError(KafkaToolError):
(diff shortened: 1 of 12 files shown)


kafka_utils/kafka_cluster_manager/cluster_info/error.py
```diff
@@ -20,11 +20,23 @@ class InvalidBrokerIdError(KafkaToolError):
     pass
 
 
+class InvalidBrokerError(KafkaToolError):
```
ecanzonieri: InvalidBrokerError seems a bit of a broad error. We can either just use assertion or rename this exception as BrokerNotInReplicationGroup error. However, this condition would happen only if there was a code bug in the algorithm itself. In these cases I usually prefer to use "assert".

kafka_utils/kafka_cluster_manager/cluster_info/rg.py
```diff
@@ -383,6 +385,34 @@ def _get_eligible_broker_pair(self, under_loaded_rg, eligible_partition):
...
+                p_name=partition.name,
+                broker=broker.id,
+            ),
+        )
+        broker.add_partition(partition)
+
+    def remove_replica(self, partition, broker_subset=None):
+        for broker in broker_subset:
+            if broker not in self.brokers:
+                raise InvalidBrokerError(
```
ecanzonieri: My comment above refer to this exception

tests/util/validation_test.py
```diff
@@ -355,6 +355,27 @@ def test_validate_plan_base_replication_factor_changed():
...
+    }
+    base_assignment = {
+        "version": 1,
+        "partitions": [
+            {"partition": 0, "topic": u't1', "replicas": [0, 2]},
+            {"partition": 0, "topic": u't2', "replicas": [0, 1]}
+        ]
+    }
+
+    # Verify validation failed
```
ecanzonieri: Is this comment correct?

tests/kafka_cluster_manager/rg_test.py
```diff
@@ -615,3 +615,30 @@ def test_rebalance_brokers_for_topic_partition_imbalance(self, create_partition)
...
+        b2 = create_broker('b2', [p10, p20])
+        b3 = create_broker('b3', [p20])
+        rg = ReplicationGroup('test_rg', set([b1, b2, b3]))
+        rg.add_replica(p10)
+
+        assert b1.partitions == set([p10, p20])
+        assert b2.partitions == set([p10, p20])
+        assert b3.partitions == set([p10, p20])
+
+    def test_remove_replica(self, create_partition):
```
ecanzonieri: Add a test where the condition broker in rg is not verified (exception or assertion is thrown).

PR #67, opened 2016-10-03
Adding option to print offset-distance

88manpreet:
E.g.: 

Cluster name: abc, consumer group: def
Topic Name: xyz
	Partition ID: 0
		High Watermark: 2488
		Low Watermark: 2484
		Current Offset: 125
		Offset Distance: 2363 (Percentage behind: 0.95%)

### DIFF
--- a/kafka_utils/kafka_consumer_manager/commands/offset_get.py
+++ b/kafka_utils/kafka_consumer_manager/commands/offset_get.py
@@ -156,8 +156,6 @@ class OffsetGet(OffsetManagerBase):
                             metadata_tuple.highmark
                     else:
                         per_behind = 0.0
-                    per_behind = 0.0
-
                     print(
                         "\t\tOffset Distance: {distance} (Percentage behind: "
                         "{per_behind:.2f}%)".format(
--- a/kafka_utils/kafka_consumer_manager/commands/offset_get.py
+++ b/kafka_utils/kafka_consumer_manager/commands/offset_get.py
@@ -60,7 +60,7 @@ class OffsetGet(OffsetManagerBase):
             help="Type of offset watermark. \"high\" represents the offset "
             "corresponding to the latest message. \"low\" represents "
             "the offset corresponding to the earliest message. \"current\" "
-            "represents the current commited offset for this consumer. "
+            "represents the current committed offset for this consumer. "
             "\"distance\" represents the offset distance b/w high-watermark "
             "and current-offset.",
             default="all"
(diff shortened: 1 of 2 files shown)


kafka_utils/kafka_consumer_manager/commands/offset_get.py
```diff
@@ -148,3 +150,16 @@ def print_output(cls, consumer_offsets_metadata, watermark_filter):
                             current=metadata_tuple.current
                         )
                     )
+                if watermark_filter == "all" or watermark_filter == "distance":
+                    if metadata_tuple.highmark > 0:
+                        per_behind = float(metadata_tuple.highmark - metadata_tuple.current) / \
```
yzernik: Would this be the ratio behind? For the percentage you need to multiply by 100.

Also, do we need the percentage behind?

kafka_utils/kafka_consumer_manager/commands/offset_get.py
```diff
@@ -148,3 +150,16 @@ def print_output(cls, consumer_offsets_metadata, watermark_filter):
...
+                            metadata_tuple.highmark
+                    else:
+                        per_behind = 0.0
+                    print(
+                        "\t\tOffset Distance: {distance} (Percentage behind: "
```
ecanzonieri: How does this look like in json format?

88manpreet: Json example: 

[
  {
    "topic": "scribe.devc.yp_page_views",
    "partition": 0,
    "current": 121,
    "highmark": 124,
    "lowmark": 124,
    "offset_distance": 3,
    "percentage_distance": 2.42
  }
]

New console output: 
Topic Name: scribe.devc.yp_page_views
	Partition ID: 0
		High Watermark: 124
		Low Watermark: 124
		Current Offset: 121
		Offset Distance: 3
		Percentage Distance: 2.42%

kafka_utils/kafka_consumer_manager/commands/offset_get.py
88manpreet: Updated.

kafka_utils/kafka_consumer_manager/commands/offset_get.py
```diff
@@ -148,3 +160,26 @@ def print_output(cls, consumer_offsets_metadata, watermark_filter):
                             current=metadata_tuple.current
                         )
                     )
+                if watermark_filter == "all" or watermark_filter == "distance":
+                    per_distance = cls.percentage_distance(metadata_tuple.highmark, metadata_tuple.current)
```
ecanzonieri: Why do we need to cast to int? Can't we cast to int once here when we invoke the function instead than performing the cast multiple times inside percentage_distance()?

88manpreet: updated

PR #70, opened 2016-10-17
offset_get ignores --storage option when checking group existence

neoeahit:
offset_get ignores --storage option when checking group existence

This will check if storage option is kafka, and in case it is, it will read list of topics form kafka and not from zookeeper. Default option is still zookeeper.

### DIFF
--- a/kafka_utils/kafka_consumer_manager/commands/offset_manager.py
+++ b/kafka_utils/kafka_consumer_manager/commands/offset_manager.py
@@ -32,25 +32,12 @@ class OffsetManagerBase(object):
         cls,
         cluster_config,
         groupid,
+        storage='zookeeper',
         fail_on_error=True,
     ):
-        topics = []
-        with ZK(cluster_config) as zk:
-            # Query zookeeper to get the list of topics that this consumer is
-            # subscribed to.
-            try:
-                topics = zk.get_my_subscribed_topics(groupid)
-            except NoNodeError:
-                if fail_on_error:
-                    print(
-                        "Error: Consumer Group ID {groupid} does not exist.".format(
-                            groupid=groupid
-                        ),
-                        file=sys.stderr
-                    )
-                    sys.exit(1)
-
-        return topics
+        if storage == 'kafka':
+            return cls.get_topics_for_group_from_kafka(cluster_config, groupid)
+        return cls.get_topics_for_group_from_zookeeper(cluster_config, groupid, fail_on_error)
 
     @classmethod
     def preprocess_args(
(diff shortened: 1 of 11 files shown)


kafka_utils/kafka_consumer_manager/util.py
```diff
@@ -113,3 +126,91 @@ def prompt_user_input(in_str):
             sys.exit(0)
         if answer == "y" or answer == "yes":
             return
+
+
+class InvalidMessageException(Exception):
+    pass
+
+
+class KafkaGroupReader:
```
neoeahit: This method was untested, in moving it around, I hv

kafka_utils/kafka_consumer_manager/commands/offset_manager.py
```diff
@@ -77,11 +79,15 @@ def preprocess_args(
                     groupid=groupid,
                 ),
             )
-        topics = cls.get_topics_from_consumer_group_id(
-            cluster_config,
-            groupid,
-            fail_on_error,
-        )
+        if storage == 'zookeeper':
```
ecanzonieri: Should this condition be inside get_topics_from_consumer_group_id?

kafka_utils/kafka_consumer_manager/commands/offset_manager.py
```diff
@@ -77,11 +79,15 @@ def preprocess_args(
...
+            topics = cls.get_topics_from_consumer_group_id(
+                cluster_config,
+                groupid,
+                fail_on_error,
+            )
+        else:
+            kafka_group_reader = KafkaGroupReader(cluster_config)
```
ecanzonieri: I think get_topics_from_consumer_group_id should be able to fetch from either Kafka or Zookeeper. We should be able to move this logic inside that function.

tests/kafka_consumer_manager/test_utils.py
```diff
@@ -0,0 +1,137 @@
+import mock
+from kafka.common import ConsumerTimeout
+
+from kafka_utils.kafka_consumer_manager.util import KafkaGroupReader
+from kafka_utils.util.offsets import PartitionOffsets
+
+
+class TestUtils(object):
```
ecanzonieri: Rename to TestKafkaGroupReader

tests/kafka_consumer_manager/test_utils.py
```diff
@@ -0,0 +1,137 @@
...
+                    0,
+                    None
+                ]
+        ):
+            kafka_group_reader.process_consumer_offset_message('test message')
+            assert kafka_group_reader.kafka_groups == {}
+
+    def test_read_groups(self):
+
+        class Mock_Consumer(object):
```
ecanzonieri: We don't really need this class in this test. We should be able to use mock_consumer.return_value.next.return_value = mock.Mock(partition=0, topic='test_topic')

tests/kafka_consumer_manager/test_utils.py
```diff
@@ -0,0 +1,137 @@
...
+                        45
+                    ]
+                ):
+                    kafka_group_reader.read_groups()
+                    assert kafka_group_reader.kafka_groups['test_group'] == {"test_topic"}
+                    assert len(kafka_group_reader.finished_partitions) == 1
+
+    def test_read_groups_with_consumer_timeout(self):
+
+        class Mock_Consumer(object):
```
ecanzonieri: Same here. We should be able to use side_effect to raise an exception

setup.py
```diff
@@ -55,6 +55,7 @@
         "requests-futures>0.9.0",
         "kafka-python<1.0.0",
         "requests<3.0.0",
+        'retrying'
```
ecanzonieri: Please refresh requirements.txt as well.

PR #74, opened 2016-11-03
Move pending assignment plan check 

lsheng:
KAFKA-2392 Move pending assignment plan check to be a precondition of cluster manager operations so it is more obvious that it could become a blocker


kafka_utils/kafka_cluster_manager/cmds/command.py
```diff
@@ -79,10 +79,6 @@ def add_subparser(self, subparsers):
     def execute_plan(self, plan, allow_rf_change=False):
         """Save proposed-plan and execute the same if requested."""
         if self.should_execute():
-            # Exit if there is an on-going reassignment
```
ecanzonieri: This change will add the check for pending plan only to the decommission command. But pretty much all commands need that check. I think we should keep it in the command base class but just move it before calling run_command and creating the cluster topology.

PR #75, opened 2016-11-04
Fix case where 2 partitions were being moved while balancing partitio…

88manpreet:
…n-count over replication-groups


kafka_utils/kafka_cluster_manager/cluster_info/cluster_topology.py
```diff
@@ -398,10 +398,13 @@ def _rebalance_groups_partition_cnt(self):
                 ))
                 # Move all possible partitions
                 for eligible_partition in eligible_partitions:
-                    over_loaded_rg.move_partition_replica(
-                        under_loaded_rg,
-                        eligible_partition,
-                    )
+                    if len(over_loaded_rg.partitions) - len(under_loaded_rg.partitions) > 1:
```
ecanzonieri: Add a comment with the rationale here please.

PR #79, opened 2016-11-25
Offline partitions check

fede1024:
Add offline partition check as subcommand in kafka-check, refactor kafka-check command and add documentation.


kafka_utils/kafka_check/commands/command.py
```diff
@@ -44,20 +44,26 @@ def run_command(self):
 
     def run(self, cluster_config, args):
         self.cluster_config = cluster_config
+        if args.broker_id == -1:
```
billyevans: I think we need to put this only for case with first_broker or controller. Because we also want to run this command without coordination, then we don't need broker_id and data_path

kafka_utils/kafka_check/main.py
```diff
@@ -106,6 +117,9 @@ def run():
     # to prevent flooding for sensu-check.
     logging.getLogger('kafka').setLevel(logging.CRITICAL)
 
+    if (args.controller_only or args.first_broker_only) and args.broker_id is None:
```
billyevans: Probably these options controller_only and first_broker_only should be mutually exclusive. We can add check for that.

kafka_utils/kafka_check/commands/offline.py
```diff
@@ -0,0 +1,66 @@
...
+from kafka_utils.kafka_check import status_code
+from kafka_utils.kafka_check.commands.command import KafkaCheckCmd
+from kafka_utils.util.metadata import get_topic_partition_metadata
+
+
+class OfflineCmd(KafkaCheckCmd):
+
+    def build_subparser(self, subparsers):
+        subparser = subparsers.add_parser(
+            'offline',
```
fede1024: It would be, but i wanted to use the same naming convention that we use for the `under_replicated` check, that doesn't have `partitions` in it.

billyevans:
+1

billyevans:
+1

PR #81, opened 2016-12-09
Offset manager single partition

fede1024:
The kafka-consumer-manager `offset_get` command, when used with offset storage set to `kafka`, reads the entire `__consumer_offsets` topic to find out the list of topics associated with a given group.
Given that `__consumer_offsets` can be several GBs, this can take quite a long time.

With this change, `offset_get` consumes only from the partition that stores the information on the specified group (kafka code [here](https://github.com/apache/kafka/blob/d092673838173d9dedbf5acf3f4e2cd8c736294f/core/src/main/scala/kafka/coordinator/GroupMetadataManager.scala)). This change should give us a 50x speed improvement on average.

### DIFF
--- a/kafka_utils/kafka_consumer_manager/util.py
+++ b/kafka_utils/kafka_consumer_manager/util.py
@@ -154,7 +154,7 @@ class KafkaGroupReader:
         self.finished_partitions = set()
         self.retry_max = 3
 
-    def read_group(self, group_id=None):
+    def read_group(self, group_id):
         partition = get_group_partition(group_id)
         return self.read_groups(partition).get(group_id, [])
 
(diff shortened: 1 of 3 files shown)


kafka_utils/kafka_consumer_manager/util.py
```diff
@@ -129,6 +130,17 @@ def prompt_user_input(in_str):
...
 
 
+def get_group_partition(group):
+    """Given a group name, return the partition number of the consumer offset
+    topic containing the data associated to that group."""
+    def java_string_hashcode(s):
+        h = 0
+        for c in s:
+            h = (31 * h + ord(c)) & 0xFFFFFFFF
+        return ((h + 0x80000000) & 0xFFFFFFFF) - 0x80000000
```
ecanzonieri: +1 to link it the kafka docs or kafka code in a comment

fede1024: I've added a comment to the external function.

PR #83, opened 2016-12-14
Fix storage options

fede1024:
The `offset_advance` command is ignoring the `storage` parameter.

After this fix, the `offset_get` command cannot be used to create the `__consumer_offsets` topic anymore, so I had to call the `set_consumer_offsets` explicitly.

### DIFF
--- a/tests/acceptance/steps/util.py
+++ b/tests/acceptance/steps/util.py
@@ -22,6 +22,7 @@ from kafka.common import LeaderNotAvailableError
 
 from kafka_utils.util import config
 from kafka_utils.util.client import KafkaToolClient
+from kafka_utils.util.offsets import set_consumer_offsets
 
 
 ZOOKEEPER_URL = 'zookeeper:2181'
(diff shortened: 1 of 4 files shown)


tests/acceptance/steps/util.py
```diff
@@ -147,6 +148,6 @@ def initialize_kafka_offsets_topic():
         return
     topic = create_random_topic(1, 1)
     produce_example_msg(topic, num_messages=1)
-    create_consumer_group(topic, 'foo')
-    call_offset_get('foo', storage='kafka')
+    kafka = KafkaToolClient(KAFKA_URL)
+    set_consumer_offsets(kafka, "group", {topic: {0: 0}}, offset_storage='kafka')
```
ecanzonieri: minor: can we use something like "fake_group" instead of "group"?

fede1024: I'm pretty sure at some point in my experiments i started generating the group id randomly. I could probably go back to that solution.

PR #86, opened 2017-01-04
Adding storage option "kafka" for multiple kafka_consumer_manger subcommands

neoeahit:
Refers internal ticket KAFKA-3663, KAFKA-3716 and KAFKA-3701.

Changes include namely:

- 1. Fixing storage option "kafka" for copy_groups, list_groups, delete_groups and rename_groups
- 2. Adding acceptance tests to test above feature.
- 3. Changing the retry from 5 to 10 in client. Since acceptance tests start from copy_groups now(with offset_storage "kafka"), they require more retrying to interact with kafka storage on the initial attempt.
- 4. Changing the offset for deleting group to 0 from -1 in nullify_offsets.


tests/kafka_consumer_manager/test_copy_group.py
```diff
@@ -47,7 +47,26 @@ def mock_kafka_info(self, topics_partitions):
...
-    def test_run(self, mock_client):
+    def test_run_with_kafka(self, mock_client):
+        topics_partitions = {
+            "topic1": [0, 1, 2],
+            "topic2": [0, 1]
+        }
+        with self.mock_kafka_info(
+            topics_partitions
+        ) as (mock_process_args, mock_user_confirm, mock_ZK), \
+            mock.patch('kafka_utils.kafka_consumer_manager.commands.'
```
ecanzonieri: Use autospec

tests/kafka_consumer_manager/test_copy_group.py
```diff
@@ -47,7 +47,26 @@ def mock_kafka_info(self, topics_partitions):
...
+                       'copy_group.get_current_consumer_offsets') \
+            as mock_get_current_consumer_offsets, \
+            mock.patch('kafka_utils.kafka_consumer_manager.commands.'
```
ecanzonieri: Use autospec

tests/kafka_consumer_manager/test_rename_group.py
```diff
@@ -47,6 +47,25 @@ def mock_kafka_info(self, topics_partitions):
...
 
+    def test_run_with_kafka(self, mock_client):
+        topics_partitions = {
+            "topic1": [0, 1, 2],
+            "topic2": [0, 1]
+        }
+        with self.mock_kafka_info(
+            topics_partitions
+        ) as (mock_process_args, mock_user_confirm, mock_ZK), \
+            mock.patch('kafka_utils.kafka_consumer_manager.commands.'
```
ecanzonieri: Use autospec

tests/kafka_consumer_manager/test_rename_group.py
```diff
@@ -47,6 +47,25 @@ def mock_kafka_info(self, topics_partitions):
...
+                       'rename_group.get_current_consumer_offsets') \
+            as mock_get_current_consumer_offsets, \
+            mock.patch('kafka_utils.kafka_consumer_manager.commands.'
```
ecanzonieri: Use autospec

tests/kafka_consumer_manager/test_rename_group.py
```diff
@@ -47,6 +47,25 @@ def mock_kafka_info(self, topics_partitions):
...
+                       'rename_group.set_consumer_offsets') as mock_set_consumer_offsets:
+            cluster_config = mock.Mock(zookeeper='some_ip')
+            args = mock.Mock(source_groupid='old_group', dest_groupid='new_group', storage='kafka')
+            RenameGroup.run(args, cluster_config)
+            assert mock_set_consumer_offsets.call_count == 2
```
ecanzonieri: How are we verifying that we actually use the kafka storage?

tests/kafka_consumer_manager/test_copy_group.py
```diff
@@ -47,7 +47,26 @@ def mock_kafka_info(self, topics_partitions):
...
+                       'copy_group.set_consumer_offsets') as mock_set_consumer_offsets:
+            cluster_config = mock.Mock(zookeeper='some_ip')
+            args = mock.Mock(source_groupid='old_group', dest_groupid='new_group', storage='kafka')
+            CopyGroup.run(args, cluster_config)
+            assert mock_set_consumer_offsets.call_count == 1
+            assert mock_get_current_consumer_offsets.call_count == 1
```
ecanzonieri: How do we verify that we actually use the Kafka storage?

kafka_utils/util/offsets.py
```diff
@@ -546,7 +546,7 @@ def set_consumer_offsets(
 def _nullify_partition_offsets(partition_offsets):
     result = {}
     for partition in partition_offsets:
-        result[partition] = -1
+        result[partition] = 0
```
ecanzonieri: We can probably add a comment here, perhaps also a link to the upstream Kafka code.

neoeahit: @fede1024 I ran some tests, before and after this change. 

When the value was set to -1, this was the output which I was getting pb - 253964
When I set it to 0, the output I am getting this is pb - 253967

I have added acceptance tests to test the same. I got this value purely by trial and error. I wasn't able to find actual acceptance tests which were testing this before.
@ecanzonieri I am trying to locate the kafka code, and I will post it once I locate it.

tests/kafka_consumer_manager/test_copy_group.py
neoeahit: This is a unit test, so I am relying on the underneath functions in offsets.py to be properly working. I have added param check to verify that set/get consumer offsets is being called with correct values. There are acceptance tests which should be covering the broader cases.

kafka_utils/kafka_consumer_manager/commands/rename_group.py
```diff
@@ -70,30 +78,84 @@ def run(cls, args, cluster_config):
...
+            offset_storage='kafka',
+        )
+        set_consumer_offsets(
+            client,
+            new_groupid,
+            copied_offsets,
+            offset_storage='kafka',
+            raise_on_error=True,
+        )
+        new_offsets = nullify_offsets(topics)
```
neoeahit: done!

tests/kafka_consumer_manager/test_rename_group.py
neoeahit: Missed it! Added! Thanks!

tests/kafka_consumer_manager/test_rename_group.py
neoeahit: Missed it! Added! Thanks!

kafka_utils/util/offsets.py
fede1024: From the pb it looks like it's doing the right thing, but i wasn't able to find any documentation of this behavior. I'm worried that it might change in other versions of Kafka. However, even this behavior will change with a new version, 0 is a valid value and it will eventually run out of retention.

PR #90, opened 2017-02-14
Upgrade kafka python

ny2ko:
Use kafka python 1.3.2 with kafka utils.


kafka_utils/kafka_consumer_manager/util.py
```diff
@@ -181,19 +183,19 @@ def read_group(self, group_id):
...
             group_id='offset_monitoring_consumer',
             bootstrap_servers=self.kafka_config.broker_list,
-            auto_offset_reset='smallest',
-            auto_commit_enable=False,
+            auto_offset_reset='earliest',
+            enable_auto_commit=False,
             consumer_timeout_ms=3000,
         )
+
+        if partition:
```
fede1024: We should be `if not None` since the partition might be 0.

yzernik: Truthy check of partition value will not work here. When partition value is zero, it evaluates to False.

yzernik: Ah @fede1024 beat me to it.

kafka_utils/util/metadata.py
```diff
@@ -12,17 +12,32 @@
...
 REPLICA_NOT_AVAILABLE_ERROR = 9
 
 
 def get_topic_partition_metadata(hosts):
-    """Returns topic-partition metadata from Kafka broker."""
-    kafka_client = KafkaClient(hosts, timeout=10)
-    return kafka_client.topic_partitions
+    """Returns topic-partition metadata from Kafka broker.
+
+    kafk-python 1.3+ doesn't include partition metadata information in
```
fede1024: Typo

kafka_utils/util/client.py
```diff
@@ -59,3 +67,86 @@ def send_offset_commit_request_kafka(
 
         return [resp if not callback else callback(resp) for resp in resps
                 if not fail_on_error or not self._raise_on_response_error(resp)]
+
+    def send_consumer_metadata_request(self, payloads=[], fail_on_error=True,
+                                       callback=None):
```
neoeahit: possibly add a docstring since this is overridden, so it might be good to say what does this do differently. Also we should probably add some tests for this, since this is changing some functionality.

kafka_utils/util/protocol.py
```diff
@@ -12,46 +12,58 @@
...
-        message.append(write_short_string(group))
-        message.append(struct.pack('>i', -1))   # ConsumerGroupGenerationId
-        message.append(write_short_string(''))  # ConsumerId
-        message.append(struct.pack('>q', -1))   # Retention time
-        message.append(struct.pack('>i', len(grouped_payloads)))
-
-        for topic, topic_payloads in grouped_payloads.items():
-            message.append(write_short_string(topic))
-            message.append(struct.pack('>i', len(topic_payloads)))
+        return kafka.protocol.commit.OffsetCommitRequest[2](
```
neoeahit: Would it make sense to add https://github.com/dpkp/kafka-python/blob/master/kafka/protocol/commit.py#L64 as a comment to explain the 2, otherwise it requires a bit of context for a new person looking at this.

kafka_utils/util/offsets.py
```diff
@@ -342,18 +342,18 @@ def _commit_offsets_to_watermark(
...
             for partition in partitions
         ]
     elif watermark == LOW_WATERMARK:
         group_offset_reqs = [
-            OffsetCommitRequest(
-                kafka_bytestring(topic), partition,
-                watermark_offsets[topic][partition].lowmark, None
+            OffsetCommitRequestPayload(
+                topic, partition,
+                watermark_offsets[topic][partition].lowmark, ''
```
neoeahit: could we say metadata='', is more descriptive and helps not to jump around in code

yzernik:
+1

kafka_utils/util/client.py
ny2ko: This is simply calling KafkaToolProtocol instead of KafkaProtocol. The real change is in KafkaToolProtocol which actually has working encode_consumer_metadata_request and decode_consumer_metadata_request methods compared to the upstream ones that don't actually work anymore. The changes we'd care about would be in kafka_utils/util/protocol.py but I can put a comment making that explicit

kafka_utils/util/protocol.py
ny2ko: I'd prefer not to have a link as a comment as the link could go stale as this file is modified

kafka_utils/util/client.py
neoeahit: @ny2ko++

neoeahit:
+1

kafka_utils/kafka_consumer_manager/util.py
```diff
@@ -181,19 +183,19 @@ def read_group(self, group_id):
...
-            auto_commit_enable=False,
+            auto_offset_reset='earliest',
+            enable_auto_commit=False,
             consumer_timeout_ms=3000,
         )
+
+        if partition is not None:
+            self.consumer.assign([TopicPartition(CONSUMER_OFFSET_TOPIC, partition)])
+        else:
+            self.consumer.subscribe([CONSUMER_OFFSET_TOPIC])
```
ecanzonieri: Let's always use manual assignment since we do not need consumer coordination here.

kafka_utils/kafka_consumer_manager/util.py
```diff
@@ -194,28 +188,21 @@ def read_groups(self, partition=None):
...
-                break
-            except (
-                    FailedPayloadsError,
-                    KafkaUnavailableError,
-                    LeaderNotAvailableError,
-                    NotLeaderForPartitionError,
-            ) as e:
-                self.retry += 1
-                self.log.warning("Got %s, retrying", e.__class__.__name__)
+            message = self.consumer.next()
```
ecanzonieri: Message is None in case of timeout. Perhaps we can just validate that condition and throw an error.

PR #95, opened 2017-03-01
Offset get sort output

fede1024:
Sorting the output of offset_get by total lag for each topic is useful in case a consumer is consuming from many topics and we want to find which one or how many of them are lagging behind by a considerable amount.

### DIFF
--- a/kafka_utils/kafka_consumer_manager/commands/offset_get.py
+++ b/kafka_utils/kafka_consumer_manager/commands/offset_get.py
@@ -75,11 +75,12 @@ class OffsetGet(OffsetManagerBase):
             "-j", "--json", action="store_true",
             help="Export data in json format."
         )
-        parser_offset_get.add_argument(
+        sort_parser = parser_offset_get.add_mutually_exclusive_group()
+        sort_parser.add_argument(
             "--sort-by-distance", action="store_true",
             help="Sort the output by increasing topic distance."
         )
-        parser_offset_get.add_argument(
+        sort_parser.add_argument(
             "--sort-by-distance-percentage", action="store_true",
             help="Sort the output by increasing topic distance percentage."
         )
@@ -87,10 +88,6 @@ class OffsetGet(OffsetManagerBase):
 
     @classmethod
     def run(cls, args, cluster_config):
-        if args.sort_by_distance and args.sort_by_distance_percentage:
-            print("Invalid sorting parameters.", file=sys.stderr)
-            sys.exit(1)
-
         # Setup the Kafka client
         client = KafkaToolClient(cluster_config.broker_list)
         client.load_metadata_for_topics()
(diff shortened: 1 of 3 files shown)


kafka_utils/kafka_consumer_manager/commands/offset_get.py
```diff
@@ -74,10 +75,22 @@ def setup_subparser(cls, subparsers):
...
+        )
+        parser_offset_get.add_argument(
+            "--sort-by-distance-percentage", action="store_true",
+            help="Sort the output by increasing topic distance percentage."
+        )
         parser_offset_get.set_defaults(command=cls.run)
 
     @classmethod
     def run(cls, args, cluster_config):
+        if args.sort_by_distance and args.sort_by_distance_percentage:
```
ecanzonieri: Prefer using https://docs.python.org/3/library/argparse.html#mutual-exclusion

kafka_utils/kafka_consumer_manager/commands/offset_get.py
```diff
@@ -140,7 +181,8 @@ def get_offsets(cls, client, group, topics_dict, storage):
     @classmethod
     def print_output(cls, consumer_offsets_metadata, watermark_filter):
         for topic, metadata_tuples in consumer_offsets_metadata.iteritems():
-            print ("Topic Name: {topic}".format(topic=topic))
+            diff_sum = sum([t.highmark - t.current for t in metadata_tuples])
+            print ("Topic Name: {topic}  Total difference: {diff}".format(topic=topic, diff=diff_sum))
```
ny2ko: I'm not sure about total difference here, can we name it something like offset difference?

kafka_utils/kafka_consumer_manager/commands/offset_get.py
fede1024: Definitely better. Thanks.

kafka_utils/kafka_consumer_manager/commands/offset_get.py
fede1024: That should be "Distance", since it's the same word used in the rest of the output. I'll change it. I think adding "offset" would be redundant since it's obvious from the rest of the output.

New output:
```
Topic Name: topic0  Total Distance: 167
        Partition ID: 0
                High Watermark: 256769105
                Low Watermark: 256456931
                Current Offset: 256769079
                Offset Distance: 26
                Percentage Distance: 0.0%
        Partition ID: 1
                High Watermark: 25124601
                Low Watermark: 24822641
                Current Offset: 25124460
                Offset Distance: 141
                Percentage Distance: 0.0%
Topic Name: topic1  Total Distance: 348373
        Partition ID: 0
                High Watermark: 46817713
                Low Watermark: 43965920
                Current Offset: 46469340
                Offset Distance: 348373
                Percentage Distance: 0.74%
```

fede1024: Let me know if this works better.

PR #97, opened 2017-03-02
Add generic prechecks

neoeahit:
Refers to internal ticket: KAFKA-4493

Please note, this is implemented in a way, to keep the rolling-restart class as generic as possible, giving us ability to add whatever prechecks we need in a pluggable way.

This commit incorporates two prechecks:
1. adding a precheck version_precheck to check for kafka package name and version before running rolling restart
2. adding a precheck config_precheck to check for configs on the broker, and only when they exist execute rolling restart

The functionality has been implemented to allow fail fast. 

example command:

kafka-rolling-restart -t test --precheck kafka_utils.kafka_rolling_restart.version_precheck --precheck-args "--package-name confluent-kafka --package-version 0.10.1.1" --precheck kafka_utils.kafka_rolling_restart.config_precheck --precheck-args "--ensure-configs auto.create.topics.enable=false"


kafka_utils/kafka_rolling_restart/precheck.py
```diff
@@ -0,0 +1,65 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import shlex
+
+
+class PrecheckFailedException(Exception):
+    pass
+
+
+class Precheck(object):
```
ecanzonieri: What about call this Task and we can then distinguish between LocalPreStopTask and RemotePreStopTask. LocalPreTask is executed locally while remote is executed on the remote broker. I'd suggest to also add RemotePostStopTask and LocalPostStopTask (executed after the stop but before the start of the Kafka broker).

This would allow enable users to do tasks such as instance type bump, instance reboot (ex. to install a new driver), etc.

kafka_utils/kafka_rolling_restart/main.py
```diff
@@ -279,6 +312,24 @@ def wait_for_stable_cluster(
         time.sleep(check_interval)
 
 
+def execute_task(checks, host):
+    """execute all the prechecks for the host
+    This trying to run(), and on getting a PrecheckFailedException executes failure().
+    It will try to run run() before exiting(incase of faiilure) to assert if
+    precheck is now valid()
+    """
+    for check in checks:
```
ecanzonieri: task in tasks

kafka_utils/kafka_rolling_restart/main.py
```diff
@@ -279,6 +312,24 @@ def wait_for_stable_cluster(
...
+        try:
+            check.run(host)
+        except TaskFailedException:
+            check.failure(host)
```
ecanzonieri: do we really need failure and success? Can we just make the Exception bubble up and fail?

kafka_utils/kafka_rolling_restart/main.py
neoeahit: I had to change it from task because of this error:

kafka_utils/kafka_rolling_restart/main.py
neoeahit: The way this has been structured is every task is executed, and upon execution, there is an outcome. A user may want to do some success or failure things on a task execution. This is why I made it this way.

Also on failure I am triggering puppet upgrade, from the task function. This allows me to have less logic in main.py and move all the update else logic in these pluggable modules.

ecanzonieri: Right, but my point is that this can be done in the run() implementation itself. The code executing a task doesn't need to know about the task specific error handling. We can call task.run() then any internal failure can be handled in the task class itself. If eventually, the exception bubbles up then rolling restart will just fail.

neoeahit: cool. Have removed the success and failure methods.

kafka_utils/kafka_rolling_restart/main.py
```diff
@@ -367,6 +412,34 @@ def validate_opts(opts, brokers_num):
...
+    :type tasks: list
+    :param task_args: list of strings to be used as args
+    :type task_args: list
+    """
+    pre_stop_task = []
+    post_stop_task = []
+    task_args_itr = task_args.__iter__()
+    tasks_classes = [PreStopTask, PostStopTask]
+
+    for func in tasks:
```
ny2ko: I think that if you use enumerate here, it gets rid of you having to use task_args_iter, alternatively you could do:

task_to_task_args = dict(zip(tasks, task_args))
for func, task_args in task_to_task_args.items():
    ....

kafka_utils/kafka_rolling_restart/main.py
```diff
@@ -367,6 +412,34 @@ def validate_opts(opts, brokers_num):
...
+def get_task_class(tasks, task_args):
+    """Reads in a list of tasks provided by the user,
+    loads the appropiate task, and returns two lists,
+    pre_stop_tasks and post_stop_tasks
+    :param tasks: list of strings locating tasks to load
+    :type tasks: list
+    :param task_args: list of strings to be used as args
+    :type task_args: list
+    """
+    pre_stop_task = []
```
ny2ko: nitpick: mind making these two plural

kafka_utils/kafka_rolling_restart/main.py
```diff
@@ -367,6 +412,34 @@ def validate_opts(opts, brokers_num):
...
+    post_stop_task = []
+    task_args_itr = task_args.__iter__()
+    tasks_classes = [PreStopTask, PostStopTask]
+
+    for func in tasks:
+        for task_class in tasks_classes:
+            imported_class = dynamic_import(func, task_class)
+            if imported_class:
+                if task_class is PreStopTask:
```
ny2ko: I think you may be looking to use if isinstance instead of is for both this and PostStopTask

kafka_utils/kafka_rolling_restart/task.py
```diff
@@ -0,0 +1,57 @@
...
+        Please note an exception from this method will completely stop the restart
+        :param host: the host on which precheck is executed on
+        :type host: string
+        """
+        raise NotImplementedError("Implemented in subclass")
+
+
+class PreStopTask(Task):
+    """Class to be used for any pre stop checks"""
+
```
ny2ko: I'm surprised this doesn't need pass. TIL

kafka_utils/kafka_rolling_restart/main.py
neoeahit: We are importing classes using dynamic_import, and I havent formed instances at this point, hence I cannot use isinstance for this check.

kafka_utils/kafka_rolling_restart/config_precheck.py
```diff
@@ -0,0 +1,91 @@
...
+from .task import PreStopTask
+from .task import TaskFailedException
+
+
+KAFKA_UPDATE_CMD = "run-puppet"
+DEFAULT_CONFIG_FILE_LOCATION = "/etc/kafka/server.properties"
+
+
+class ConfigPrecheck(PreStopTask):
+    """Class to check for kafka configs"""
```
ecanzonieri: We should make clear that this task will try to run puppet when the configs are missing. I think it's better to move this file in our internal repository as its content doesn't seem very relevant for the open source project.

kafka_utils/kafka_rolling_restart/main.py
```diff
@@ -279,6 +304,16 @@ def wait_for_stable_cluster(
         time.sleep(check_interval)
 
 
+def execute_task(tasks, host):
+    """execute all the prechecks for the host
+    This trying to run(), and on getting a PrecheckFailedException executes failure().
```
ecanzonieri: Fix docs

kafka_utils/kafka_rolling_restart/main.py
```diff
@@ -314,11 +351,16 @@ def execute_rolling_restart(
     :type skip: integer
     :param verbose: print commend execution information
     :type verbose: bool
+    :param pre_stop_task: a list of prechecks to execute before running stop
```
ecanzonieri: prechecks? Should we fix this docs?

kafka_utils/kafka_rolling_restart/version_precheck.py
```diff
@@ -0,0 +1,88 @@
...
+from fabric.api import task
+
+from .task import PreStopTask
+from .task import TaskFailedException
+
+CHECK_KAFKA_PACKAGE = "dpkg -s {} | grep '^Version: {}'"
+KAFKA_UPDATE_CMD = "run-puppet"
+
+
+class VersionPrecheck(PreStopTask):
```
ecanzonieri: Same as above. We should probably also rename this to CheckVersionOrUpgrade because VersionPrecheck doesn't imply that a change (upgrade) is performed when the version is not matching.

neoeahit: Good point. Il include this change. I have moved this file to our internal repo and would add those changes there.

PR #99, opened 2017-03-07
Add global event

neoeahit:
Refers to internal ticket KAFKA-4276, and provides an ability to add global events, which be run before the start and end of rolling restart. 
The aim internally is to tie these to signalfx, so that we can view when these rolling restart activities are performed

### DIFF
--- a/kafka_utils/kafka_rolling_restart/main.py
+++ b/kafka_utils/kafka_rolling_restart/main.py
@@ -145,7 +145,7 @@ def parse_opts():
     parser.add_argument(
         '--global-event',
         type=str,
-        help='Module containing an implementation of Task.'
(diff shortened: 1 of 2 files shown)


kafka_utils/kafka_rolling_restart/main.py
```diff
@@ -141,6 +142,14 @@ def parse_opts():
         action='append',
         help='Arguements which are needed by the task(prestoptask or poststoptask).'
     )
+    parser.add_argument(
+        '--global-event',
```
ny2ko: Can we name this something other than global event? Maybe --perform-event?

neoeahit:
Closing this review after having a chat internally with @ecanzonieri

PR #102, opened 2017-03-29
Add handy command line print for offline partitions check with -v option

billyevans:
It should simplify investigation of offline partitions for Kafka cluster.

### DIFF
--- a/kafka_utils/kafka_check/commands/offline.py
+++ b/kafka_utils/kafka_check/commands/offline.py
@@ -14,6 +14,8 @@
 # limitations under the License.
 from __future__ import absolute_import
 
+import sys
+
 from kafka_utils.kafka_check import status_code
 from kafka_utils.kafka_check.commands.command import KafkaCheckCmd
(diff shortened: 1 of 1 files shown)


kafka_utils/kafka_check/commands/offline.py
```diff
@@ -62,6 +64,10 @@ def _prepare_output(partitions, verbose):
                 for (topic, partition) in partitions
             )
             out['verbose'] = "Partitions:\n" + "\n".join(lines)
+        else:
+            cmdline = sys.argv[:]
+            cmdline.insert(1, '-v')
+            out['message'] += '\nTo see all offline partitions run: ' + ' '.join(cmdline)
```
ecanzonieri: Perhaps we could simply print "use --verbose (-v) to see all offline partitions"

billyevans: I think sometimes it could take some time to find the same keys to run the command. I believe this will save some time for us.

88manpreet:
+1

PR #105, opened 2017-04-04
Add acceptance test for unsubscribe command.

yzernik:


tests/acceptance/steps/unsubscribe_topics.py
```diff
@@ -43,3 +45,23 @@ def step_impl4(context):
     with ZK(cluster_config) as zk:
         offsets = zk.get_group_offsets(context.group)
     assert context.topic not in offsets
+
+
+@when('we call the unsubscribe_topics command with kafka storage')
+def step_impl2_kafka(context):
+    call_unsubscribe_topics(context.group, 'kafka')
```
neoeahit: since kafka is the default storage, to we should not pass storage option. This would be a more real scenario. What do you think?

tests/acceptance/steps/unsubscribe_topics.py
```diff
@@ -43,3 +45,23 @@ def step_impl4(context):
...
+def step_impl4_kafka(context):
+    cluster_config = get_cluster_config()
+    client = KafkaToolClient(cluster_config.broker_list)
+    client.load_metadata_for_topics()
+
+    offsets = get_current_consumer_offsets(
+        client,
+        context.group,
+        [context.topic],
+        offset_storage='kafka',
```
neoeahit: nit: this is the default option, we can probably remove it

tests/acceptance/steps/unsubscribe_topics.py
```diff
@@ -43,3 +45,23 @@ def step_impl4(context):
...
+    )
+    assert offsets[context.topic] == {0: 0}
```
neoeahit: If a topic is unsubscribed its offsets should not be present in the returned response.

yzernik: Good point. We have to decide what it means to be unsubscribed. Having consumer offset at zero may or may not be unsubscribed.

Also, this test should use offset_get command to fetch the offsets.

PR #113, opened 2017-05-09
Don't limit genetic algorithm plan exploration

ecanzonieri:
This change will make genetic rebalancer complete the full exploration when performing a rebalance. Completing the full exploration is important to reach a score value that is above the threshold.
The plan is then reduced right before the assignment.

### DIFF
--- a/tests/kafka_cluster_manager/genetic_balancer_test.py
+++ b/tests/kafka_cluster_manager/genetic_balancer_test.py
@@ -144,13 +144,17 @@ class TestGeneticBalancer(object):
         """
         assert self.move_partition_valid(5, 1, 3)
 
-    def test_move_partition_too_many_movements(self):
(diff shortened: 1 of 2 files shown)


tests/kafka_cluster_manager/genetic_balancer_test.py
```diff
@@ -144,13 +144,17 @@ def test_move_partition_move_rg(self):
         """
         assert self.move_partition_valid(5, 1, 3)
 
-    def test_move_partition_too_many_movements(self):
+    def test_move_partition_ignores_max_partition_movements(self):
         """Test _move_partition when the partition movement limit has been
         reached.
 
-        Moving any partition is invalid because max_partition_movements is 0.
+        Moving any partition is still valide despite max_partition_movements is 0.
```
fede1024: s/valide/valid

PR #117, opened 2017-05-26
Add python3.4+ support

kennydo:
The bulk of this change is automated changes made by https://python-modernize.readthedocs.io/en/latest/ to use `six` (all in this one commit https://github.com/kennydo/kafka-utils/commit/9b537d83b2d862ab03cca2b5afd30024b62d4cb4)

I then make a few manual tweaks, and then made additional changes:
- I added the deadsnakes repo to the Dockerfiles so we can install packages for more python versions
- I installed `tox-travis` so that each Travis build job only triggers the appropriate Python tox env (the build job for Python 3.5 only runs the stuff for py35, and so on)
- I swapped the `Fabric` dependency with `Fabric3`, which is a drop-in replacement

For what it's worth, I had no problems using this on my local Python 3.6 environment:
```
$ python3.6 -m venv venv36
$ source venv36/bin/activate
(venv36) $ python setup.py install
(venv36) $ kafka-cluster-manager --cluster-type production --cluster-name main stats
INFO:kazoo.client:Connecting to zookeeper.imouto:2181
INFO:kazoo.client:Zookeeper connection established, state: CONNECTED
INFO:kafka-zookeeper-manager:Fetching current cluster-topology from Zookeeper...
Replication Group | Extra Replica Count
------------------+--------------------
None              | 0
[...]
``` 

Issues:
- I (think I) pass in all in the environment variables used by `tox-travis`, but the integration tests always use Python 2.7 instead of the Python version for that build job. Have you had this issue before? The unit tests all use the right python version. This seems to just affect the integration tests (maybe because something isn't passing through to the docker container?).


kennydo:
Ah-hah! I've figured out the issue with the acceptance tests. I had to add the `py{27,34,35,36}` factors into the testenv for the acceptance task. I hit the same issue described in https://github.com/tox-dev/tox/issues/465, so I followed the solution posted in that thread.

I refactored the tox testenvs so now `py{27,34,35,36}-kafka{8,9,10}-dockeritest` kick off the appropriate `py{27,34,35,36}-acceptance` (by passing in the Python version factor as an environment var into the docker container.

You can see in the Travis output that the acceptance tests are now running with the desired Python version per build job.

jparkie:
You might want to add something like https://github.com/Yelp/data_pipeline/blob/master/.pre-commit-config.yaml#L26-L29 in the .pre-commit-config.yaml.

kennydo:
I'm in favor of adding those to the pre-commit config, but I'm hesitant to mix that into this PR that already touches 100+ files.

ecanzonieri:
@kennydo that sounds good, there are already too many changes in this pr. Let's add the pre-commit-config changes in a separate branch.

kafka_utils/kafka_cluster_manager/cmds/rebalance.py
```diff
@@ -178,25 +183,25 @@ def run_command(self, cluster_topology, cluster_balancer):
...
+            list(cluster_topology.partitions.values())
         )
 
         if self.args.show_stats:
             display_cluster_topology_stats(cluster_topology, base_assignment)
             if base_score is not None and score is not None:
-                print('\nScore before: %f' % base_score)
-                print('Score after:  %f' % score)
-                print('Score improvement: %f' % (score - base_score))
+                print(('\nScore before: %f' % base_score))
```
jparkie: There are double parentheses.

ecanzonieri:
@kennydo sorry for the delay reviewing this pr. We have an urgent fix #118 to ship before merging the python3 support.

It would be great if you could rebase your branch on top of that fix once merged to master. I'll then run some additional manual testing on your branch for the tools that do not have integration tests and merge your pr afterwards.

kennydo:
No worries! I've rebased this branch.

tests/acceptance/steps/util.py
```diff
@@ -66,20 +84,31 @@ def delete_topic(topic_name):
 
 def call_cmd(cmd):
     output = ''
+
+    communicate_input = b'y' if six.PY3 else 'y'
```
jparkie: `b'y'` should be both py2 and py3 compatible: http://python-future.org/compatible_idioms.html#byte-string-literals

tests/kafka_cluster_manager/partition_count_balancer_test.py
```diff
@@ -328,15 +332,15 @@ def test__rebalance_groups_partition_cnt_case5(
         cb._rebalance_groups_partition_cnt()
 
         # Assert partition is moved from rg1 only
-        print(ct.assignment)
+        print((ct.assignment))
```
jparkie: You don't need double parentheses.

tests/util/config_test.py
```diff
@@ -173,25 +172,18 @@ def get_fake_yaml(path):
         else:
             return MOCK_YAML_2
 
-    with contextlib.nested(
```
jparkie: Thanks! :)

kafka_utils/kafka_consumer_manager/util.py
```diff
@@ -250,22 +255,22 @@ def remove_partition_from_consumer(self, partition):
...
-            value = bytearray(message.value)
+            value = message.value
             ((value_schema,), cur) = relative_unpack(b'>h', value, 0)
             if value_schema not in [0, 1]:
                 raise InvalidMessageException()  # Unrecognized message value
             ((offset,), cur) = relative_unpack(b'>q', value, cur)
         else:
             offset = None  # Offset was deleted
-        return str(group), str(topic), partition, offset
+        return group.decode(), topic.decode(), partition, offset
```
kennydo: The reason for the change here is that `str(b'123') == "b'123'"` in py3 (note the "b" in the string), but `str(b'123') == '123'` in py2, whereas `.decode()` does the right thing in both environments.

I find that `_tostr` function to be more confusing than enlightening, but I think that's a matter of preference.

tests/acceptance/steps/util.py
kennydo: Oh, good point! Will update accordingly.

tests/acceptance/steps/min_isr.py
```diff
@@ -37,7 +39,10 @@ def set_min_isr(topic, min_isr):
     cluster_config = get_cluster_config()
     with ZK(cluster_config) as zk:
         config = zk.get_topic_config(topic)
-        config['config'] = {ISR_CONF_NAME: str(min_isr)}
+
+        serialized_min_isr = str(min_isr)
```
kennydo: I think I made this change for another thing that is no longer relevent. Undone!

tests/kafka_cluster_manager/partition_count_balancer_test.py
```diff
@@ -332,7 +332,7 @@ def test__rebalance_groups_partition_cnt_case5(
         cb._rebalance_groups_partition_cnt()
 
         # Assert partition is moved from rg1 only
-        print((ct.assignment))
+        print(ct.assignment)
```
ecanzonieri: hmm, I don't think we want to keep a print in the test :)

tests/acceptance/steps/config_update.py
```diff
@@ -12,15 +12,17 @@
...
 # limitations under the License.
+from __future__ import absolute_import
+
 import time
 
 from behave import then
 from behave import when
 from kafka.errors import MessageSizeTooLargeError
+from util import get_cluster_config
+from util import produce_example_msg
```
ecanzonieri: Any specific reason for this import reordering? These are local relative import I think they should stay as they were before, right?

kafka_utils/util/zookeeper.py
```diff
@@ -39,6 +39,15 @@ def load_json(data_str):
         return json.loads(data_str)
 
 
+def dump_json(obj):
```
ecanzonieri: What about we move this function in a separate module so we do not need to re-implement in the test?

tests/util/zookeeper_test.py
```diff
@@ -23,6 +23,15 @@
 from kafka_utils.util.zookeeper import ZK
 
 
+def dump_to_json_and_encode(obj):
```
ecanzonieri: Commented above. We could move this function in the kafka_utils/util/utils.py and implement it only once.

kafka_utils/kafka_cluster_manager/cmds/command.py
```diff
@@ -200,10 +200,6 @@ def get_reduced_assignment(
             if set(replica) != set(new_assignment[t_p])
         ]
 
-        if (leaders_changes <= max_leader_only_changes and
-                partition_change_count <= max_partition_movements):
-            return new_assignment
```
ecanzonieri: Why did we remove this?

tests/acceptance/steps/watermark_get.py
```diff
@@ -16,8 +16,7 @@
 
 from behave import then
 from behave import when
-
-from .util import call_watermark_get
+from util import call_watermark_get
```
ecanzonieri: Shouldn't this be relative import?

kennydo: I got a weird "__name__ not in globals python" error when I kept it as a relative import, so I've updated all these imports to use the absolute "steps.util" path

kafka_utils/kafka_cluster_manager/cmds/command.py
kennydo: `leaders_changes` and `partition_change_count` are lists, but they're being compared to integers.

In py3, this is no longer valid. When I changed them to compare the lengths of the list to the integers, it no longer worked, so I just removed them. The tests still passed 🤔

ecanzonieri: I think tests pass because that's just a small optimization. I'm ok with removing it.

tests/acceptance/steps/watermark_get.py
ecanzonieri: That's weird. I'll merge your branch and take a look at that. It's mostly a nitpick.

ecanzonieri:
Thanks for this pr.

PR #118, opened 2017-06-02
Fix kafka topic config setter

neoeahit:
Refers internal ticket: KAFKA-5160

Currently this tool is broken, and this PR fixes this for both kafka 9 and kafka 10

### DIFF
--- a/kafka_utils/util/zookeeper.py
+++ b/kafka_utils/util/zookeeper.py
@@ -122,7 +122,7 @@ class ZK:
             raise e
         return config_data
 
-    def set_topic_config(self, topic, value):
+    def set_topic_config(self, topic, value, kafka_version=(0, 10, )):
         """Set configuration information for specified topic.
 
         :rtype : dict of new configuration"""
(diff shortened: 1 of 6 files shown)


kafka_utils/util/zookeeper.py
```diff
@@ -134,9 +134,25 @@ def set_topic_config(self, topic, value):
...
+                    "entity_type": "topics",
+                    "entity_name": topic
+                })
+            elif version == 10:
+                # https://github.com/apache/kafka/blob/0.10.2.1/
+                #     core/src/main/scala/kafka/admin/AdminUtils.scala#L574
+                change_node = json.dumps({
+                    "version": 2,
+                    "entity_path": "topics/" + topic,
+                })
```
ecanzonieri: This is missing an "else". What do we do if the version is neither 9 nor 10?

neoeahit: added an assertion to tackle that case

kafka_utils/util/zookeeper.py
```diff
@@ -134,9 +141,29 @@ def set_topic_config(self, topic, value):
                 config_data
             )
             # Create change
+            version = kafka_version[1]
+
+            # this feature is supported in kafka 9 and kafka 10
+            assert version in [9, 10], "Feature supported with kafka 9 and kafka 10"
```
ecanzonieri: Let's use a set here: {9, 10}. It's doesn't make any difference in this case, but it's a bit nicer.

tests/util/zookeeper_test.py
```diff
@@ -203,7 +203,7 @@ def test_set_topic_config(self, mock_client):
 
                 expected_create_call = mock.call(
                     '/config/changes/config_change_',
-                    "some_topic",
+                    json.dumps({"entity_path": "topics/some_topic", "version": 2}),
```
ecanzonieri: I think we should have a unit test for 0.9 as well.

neoeahit: added

kafka_utils/util/zookeeper.py
neoeahit: Discussed offline, and added ()

PR #121, opened 2017-06-08
Remove Fabric as a dependency for py3 compatibility

jparkie:
Refers to internal ticket: DATAPIPE-3049, KAFKA-5165.


kafka_utils/util/ssh.py
```diff
@@ -0,0 +1,158 @@
...
+        :param command: the command to execute
+        :type command: str
+        :param bufsize: interpreted the same way as by the built-in C{file()} function in python
+        :type bufsize: int
+        :returns the stdin, stdout, and stderr of the executing command
+        :rtype: tuple(L{ChannelFile}, L{ChannelFile}, L{ChannelFile})
+
+        :raises SSHException: if the server fails to execute the command
+        """
+        new_command = "sudo -S -p 'sudo password:' {0}".format(command)
```
ecanzonieri: Do we really need "-S -p 'sudo password:'" if yes can we add a comment to explain why?

jparkie: I checked https://github.com/fabric/fabric/blob/1.13/sites/docs/usage/env.rst#sudo_prefix. Seems like we don't really need it. What do you recommend?

ecanzonieri: I'd remove it then. @fede1024 any objection?

jparkie: @ecanzonieri Should I just replace sudo_command and just use exec_command? As long as the connection is sudoable, we can just prefix the commands with sudo.

ecanzonieri: I think having a function to execute sudo commands is more clean from an interface point of view. I would suggest to keep it.

jparkie:
@ecanzonieri I can't manually test the changes, so @fede1024 helped me test them. I think removing the `-S -p` should be fine.

fede1024:
I like the current (new) solution, with the separate sudo function and no `-S -p`. The command should be semantically identical so I don't think it's worth testing again.

ecanzonieri:
This looks good to me, but I'd wait for #117 to be merged first to reduce the number of conflicts.

ecanzonieri:
@jparkie I just merged #117 please pull master, there should be few merge conflicts to solve and then I'll merge this one as well.

PR #126, opened 2017-07-26
Add support for partition and leader count in generic balancer

ecanzonieri:
Consider the partition count CV when computing the final score. This change introduces a new criteria when rebalancing.

If the cluster is already balanced by only broker weight, the movement of the smallest partitions (smaller impact on broker weight) is implicitly preferred to improve the partition count criteria.
In clusters with lots of empty partitions this change improves the overall utilization.
 
Internal ticket KAFKA-5777

### DIFF
--- a/kafka_utils/kafka_cluster_manager/cluster_info/genetic_balancer.py
+++ b/kafka_utils/kafka_cluster_manager/cluster_info/genetic_balancer.py
@@ -49,10 +49,10 @@ DEFAULT_MAX_EXPLORATION = 10000
 # In practice, overall weight is more important than leader weight which is
 # more important than topic-broker imbalance so different weights are used
 # to adjust for this.
-DEFAULT_PARTITION_WEIGHT_CV_SCORE_WEIGHT = 0.6
-DEFAULT_LEADER_WEIGHT_CV_SCORE_WEIGHT = 0.3
-DEFAULT_TOPIC_BROKER_IMBALANCE_SCORE_WEIGHT = 0.2
-DEFAULT_BROKER_PARTITION_COUNT_SCORE_WEIGHT = 0.2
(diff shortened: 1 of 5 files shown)


kafka_utils/kafka_cluster_manager/cluster_info/genetic_balancer.py
```diff
@@ -564,6 +566,8 @@ def _score(self, state, score_movement=True):
                 (1 - state.broker_leader_weight_cv / sqrt(len(state.brokers)))
             score += self.args.topic_broker_imbalance_score_weight * \
                 (1 - state.weighted_topic_broker_imbalance)
+            score += self.args.broker_partition_count_score_weight * \
+                (1 - state.broker_partition_count_cv)
```
fede1024: Should we normalise this one as well?

ecanzonieri: Good catch. I fixed it. I also adjusted the default weights to keep the score between 0 and 1.

kafka_utils/kafka_cluster_manager/cluster_info/stats.py
```diff
@@ -51,11 +51,11 @@ def standard_deviation(data, data_mean=None, data_variance=None):
 def coefficient_of_variation(data, data_mean=None, data_stdev=None):
     """Return the coefficient of variation (CV) of a sequence of numbers.
     :param data_mean: Precomputed mean of the sequence.
-    :param data_standard_deviation: Precomputed standard_deviation of the
+    :param data_standard_deviation_p: Precomputed standard_deviation_p of the
```
fede1024: s/data_standard_deviation_p/data_stdev

ecanzonieri:
In the last pr I added support for leader count during the genetic rebalance. Any of the supported criteria can be disabled by setting it's weight to zero using the --balancer-args cli option.

PR #130, opened 2017-08-08
KAFKA-1497 - Remove docker network after acceptance tests

fede1024:
This change will configure tox to remove the docker network after every acceptance test, cleaning up the environment.


fede1024:
I also had to fix a couple extra things to make the tests happy again:
- fix flake8 complaints
- fix download scripts for old kafka versions (we should really deprecate them)

88manpreet:
+1

PR #136, opened 2017-08-31
added options to specify custom start and stop command

### LINKED ISSUE #135: Support other commands to start and stop kafka
For my setup, I have kafka running within Docker, so stopping and starting kafka isn't as straightforward as `sudo service kafka stop`. 

I'm going to submit a PR to support other options to start and stop kafka. This is not dangerous in my opinion, because it the power user already has ssh access to the Kafka Host.

simplesteph:
this fixes #135

### DIFF
--- a/kafka_utils/kafka_rolling_restart/main.py
+++ b/kafka_utils/kafka_rolling_restart/main.py
@@ -40,9 +40,6 @@ from kafka_utils.util.utils import dynamic_import
 from kafka_utils.util.zookeeper import ZK
 
 
-STOP_COMMAND = "service kafka stop"
-START_COMMAND = "service kafka start"
-
 UNDER_REPL_KEY = "kafka.server:name=UnderReplicatedPartitions,type=ReplicaManager/Value"
 
 DEFAULT_CHECK_INTERVAL_SECS = 10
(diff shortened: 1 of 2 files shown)


kafka_utils/kafka_rolling_restart/main.py
```diff
@@ -142,6 +141,20 @@ def parse_opts():
         action='append',
         help='Arguements which are needed by the task(prestoptask or poststoptask).'
     )
+    parser.add_argument(
+        '--start-command',
+        type=str,
+        help=('Override start command for kafka (do not include sudo)'
+              'Default: %(default)'),
```
ecanzonieri: 's' missing after %(default). It should be %(default)s

kafka_utils/kafka_rolling_restart/main.py
```diff
@@ -142,6 +141,20 @@ def parse_opts():
...
+        default=DEFAULT_START_COMMAND,
+    )
+    parser.add_argument(
+        '--stop-command',
+        type=str,
+        help=('Override stop command for kafka (do not include sudo)'
+              'Default: %(default)'),
```
ecanzonieri: 's' missing after %(default). It should be %(default)s

simplesteph:
@ecanzonieri thanks for the feedback. Fixed it!

PR #143, opened 2017-09-26
allowed for hostname to change its IP

### LINKED ISSUE #142: Feature: Allow for Hosts to change IP (kafka-rolling-restart)
I have my Kafka cluster in the cloud, and I have two kinds of reboot:
- A soft reboot, which is just restarting my brokers
- A hard reboot, which terminates EC2 instances, and wait for new ones to come up (using a task, thanks for the doc!)

With the hard reboot, because the IP changes and the connection is terminated, the ssh `Connection` object will fail on the `start_kafka` command. 

So we need a new SSH connection to be instantiated before `start_kafka` command is executed to ensure we execute it against the right host.

simplesteph:
this fixes #142


simplesteph:
re-test please

PR #138, opened 2017-09-06
added openssh config to kafka utils

simplesteph:
this fixes #137

PR #145, opened 2017-11-10
Use user's local ssh config

xiaochuanyu:
This attempts to parse user's local ssh config and make use of it.
Sample ssh config where this applies:
```
Host 1.2.3.4
   User ec2-user
   IdentityFile ~/.ssh/key.pem
   StrictHostKeyChecking no
   UserKnownHostsFile /dev/null
```
This allows you to setup different SSH configs for different hosts and still use the tool.


simplesteph:
@xiaochuanyu I did something similar here: https://github.com/Yelp/kafka-utils/pull/138/files
It may be worth refactoring your code and isolate the ssh client construction method for all the other components to benefit from it

PR #148, opened 2017-12-15
Improve rolling restart

TheyDroppedMe:
Added checks after broker stop to read BrokerState mbean from jolokia before starting the broker back up. While this could be implemented as a post-stop-task, I feel that this behavior is necessary to help the tool be more generic. The default behavior does not work for non-blocking init scripts as a clean broker shutdown takes time, resulting in the rolling restart tool failing, as it tries to start the broker back up immediately after shutdown, which fails and leaves a broker down. 

This is how I made the changes to work in our environment, but some things might need to be changed or reordered to fit better into the tool generically. Open to any and all suggestions.

### DIFF
--- a/kafka_utils/kafka_rolling_restart/main.py
+++ b/kafka_utils/kafka_rolling_restart/main.py
@@ -330,7 +330,7 @@ def wait_for_broker_shutdown(host, jolokia_port, jolokia_prefix, stop_check_inte
             print("Broker is stopped")
             return
         else:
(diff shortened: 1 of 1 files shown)


fede1024:
The script is designed to have a blocking "stop" command, which blocks until the process is actually terminated (and it has been definitely used on a lot of "real world" clusters :) ). You can see an example of init.d conf script [here](https://github.com/Yelp/puppet-kafka/blob/master/templates/kafka.conf.erb).

PR #154, opened 2018-02-01
Fetch creation time of topic and partition.

dongw-yelp:
Fetch creation time of topic and partition in the function "get_topics".

In order to do this, also changed the _fetch_partition_state  function to _fetch_partition_info, to return the creation time with the state information rather than only state information. Test added.

### DIFF
--- a/kafka_utils/util/zookeeper.py
+++ b/kafka_utils/util/zookeeper.py
@@ -200,6 +200,7 @@ class ZK:
         Topic-data format with fetch_partition_state as True:-
         topic_data = {
             'version': 1,
+            'ctime': <timestamp>,
             'partitions': {
                 <p_id>:{
                     replicas: [<broker_id>, <broker_id>, ...],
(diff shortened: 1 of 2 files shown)


88manpreet:
Please fix the failing tests.

kafka_utils/util/zookeeper.py
```diff
@@ -208,11 +209,13 @@ def get_topics(
...
                     version: 1,
                     leader: <broker-id>,
+                    ctime: <timestamp>,
+                }
             }
         }
-        Note: By default we also fetch partition-state which results in
-        accessing the zookeeper twice. If just partition-replica information is
-        required fetch_partition_state should be set to False.
+        Note: By default we also fetch partition-state which results.
```
88manpreet: Nit: The sentence is incomplete.

kafka_utils/util/zookeeper.py
```diff
@@ -234,9 +238,11 @@ def get_topics(
             partitions_data = {}
             for p_id, replicas in six.iteritems(topic_data['partitions']):
                 partitions_data[p_id] = {}
+                # Fetch partition-info from zookeeper
+                partition_info = self._fetch_partition_info(topic_id, p_id)
```
88manpreet: Looks like we will be fetching partition state on every call of get_topics which can slow-down the call. We should only fetch-partition-state when required.

PR #155, opened 2018-02-26
List unavailable brokers in case replicas are unavailable

88manpreet:
This change checks the list of replicas from zookeeper, and compare it with partition-metadata when error 9 (REPLICA_NOT_AVAILABLE_ERROR) is encountered, to get the list of unavailable brokers.

### DIFF
--- a/kafka_utils/kafka_check/commands/replica_unavailability.py
+++ b/kafka_utils/kafka_check/commands/replica_unavailability.py
@@ -34,17 +34,23 @@ class ReplicaUnavailabilityCmd(KafkaCheckCmd):
     def run_command(self):
         """replica_unavailability command, checks number of replicas not available
         for communication over all brokers in the Kafka cluster."""
-        replica_unavailability = get_topic_partition_with_error(
+        fetch_unavailable_brokers = True
+        result = get_topic_partition_with_error(
             self.cluster_config,
             REPLICA_NOT_AVAILABLE_ERROR,
+            fetch_unavailable_brokers=fetch_unavailable_brokers,
         )
+        if fetch_unavailable_brokers:
+            replica_unavailability, unavailable_brokers = result
+        else:
+            replica_unavailability = result
 
         errcode = status_code.OK if not replica_unavailability else status_code.CRITICAL
-        out = _prepare_output(replica_unavailability, self.args.verbose)
+        out = _prepare_output(replica_unavailability, unavailable_brokers, self.args.verbose)
         return errcode, out
 
 
-def _prepare_output(partitions, verbose):
+def _prepare_output(partitions, unavailable_brokers, verbose):
     """Returns dict with 'raw' and 'message' keys filled."""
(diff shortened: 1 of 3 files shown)


kafka_utils/util/metadata.py
```diff
@@ -42,16 +43,27 @@ def get_topic_partition_metadata(hosts):
     return topic_partitions
 
 
-def get_topic_partition_with_error(cluster_config, error):
+def get_topic_partition_with_error(cluster_config, error, get_unavailable_brokers=False):
```
ny2ko: Could we use a better name e.g. unavailable_brokers? Get usually is used for method names

kafka_utils/kafka_check/commands/replica_unavailability.py
```diff
@@ -34,17 +34,18 @@ def build_subparser(self, subparsers):
     def run_command(self):
         """replica_unavailability command, checks number of replicas not available
         for communication over all brokers in the Kafka cluster."""
-        replica_unavailability = get_topic_partition_with_error(
+        replica_unavailability, unavailable_brokers = get_topic_partition_with_error(
```
yzernik: Will it fail to unpack the tuple when get_unavailable_brokers=False ?

kafka_utils/util/metadata.py
```diff
@@ -42,16 +43,27 @@ def get_topic_partition_metadata(hosts):
...
+    if required."""
 
     metadata = get_topic_partition_metadata(cluster_config.broker_list)
     affected_partitions = set()
+    if fetch_unavailable_brokers:
+        unavailable_brokers = set([])
     for partitions in metadata.values():
         for partition_metadata in partitions.values():
             if int(partition_metadata.error) == error:
+                if fetch_unavailable_brokers:
```
yzernik: Can we have a separate function to get the unavailable brokers for a given partition_metadata and zk connection.

Also we can open the ZK connection outside of the loop so we don't keep creating new connections.

yzernik: Could be something like

```
def get_unavailable_brokers(partition_metadata, zk):
```

kafka_utils/util/metadata.py
```diff
@@ -43,6 +43,14 @@ def get_topic_partition_metadata(hosts):
     return topic_partitions
 
 
+def get_unavailable_brokers(zk, partition_metadata, unavailable_brokers):
```
yzernik: I think this would be simpler if "get_unavailable_brokers" does not take the unavailable_brokers parameter.

It could be something like this:

```
unavailable_brokers.update(
    get_unavailable_brokers(zk, partition_metadata)
)
```

yzernik: Or equivalent:

```
unavailable_brokers |= get_unavailable_brokers(zk, partition_metadata)
```

yzernik:
lgtm

PR #156, opened 2018-03-27
Add --topics option to kafka_consumer_manager unsuscribe_topics cmd

poros:
The unsuscribe_topics command of kafka_consumer_manager is now able to
unsuscribe only a subset of topics among all the ones consumed by the
consumer group. The --topics is mutually exclusive with the --topic
option and consequently with --partitions.

This is my first PR for this project, so please let me know whether I have to call in more reviewers and/or how to test my change manually (for now I only relied on unit tests).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,7 @@ pip-delete-this-directory.txt
 .cache
 nosetests.xml
 coverage.xml
+.pytest_cache/
 
 # Sphinx documentation
 docs/_build/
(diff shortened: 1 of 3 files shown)


poros:
Ah, I missed the fact that running `make test` doesn't run all the tests, but just the unit tests. So now I have made an acceptance test failing :/
I'll look into that and update the review.

poros:
I fixed the bug and amended the commit. All tests should work this time.

88manpreet:
+1 Nice test-coverage! This can be tested locally inside yelp-kafka-tool package.

poros:
Just added a whitespace change in one of the error strings. Merging as soon as the tests finish.

PR #161, opened 2018-05-11
Handle missing topics in `/config/topics` ZK path

yemyatthein:
If a legitimate topic name is missing in the `/config/topics` path, it should return a default/empty dict rather than raising an exception which should occur only if the topic is non-existent. The topic can be missing in the config path if it is created Kafka version prior to 0.8.1.


yemyatthein:
@poros Could you have a look at this please?

PR #163, opened 2018-05-11
Fix delete_group for zookeeper, make offset_get description more accu…

dongw-yelp:
For kafka-consumer-manager command, when there exists the directory consumers/{consumer group id} but not consumers/{consumer group id}/offsets in the zookeeper:

Fix the offset_get sub command saying consumer group does not exist issue.

Fix the delete_group sub command saying the consumer group does not exist issue , the expected behavior should be delete the group.

### DIFF
--- a/kafka_utils/kafka_consumer_manager/commands/offset_manager.py
+++ b/kafka_utils/kafka_consumer_manager/commands/offset_manager.py
@@ -147,14 +147,22 @@ class OffsetManagerBase(object):
             try:
                 topics = zk.get_my_subscribed_topics(groupid)
             except NoNodeError:
-                if fail_on_error:
+                if groupid in zk.get_children("/consumers"):
                     print(
-                        "Error: Consumer Group ID {groupid} does not exist.".format(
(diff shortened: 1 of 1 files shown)


kafka_utils/kafka_consumer_manager/commands/offset_manager.py
```diff
@@ -148,12 +148,20 @@ def get_topics_for_group_from_zookeeper(
                 topics = zk.get_my_subscribed_topics(groupid)
             except NoNodeError:
                 if fail_on_error:
-                    print(
-                        "Error: Consumer Group ID {groupid} does not exist.".format(
-                            groupid=groupid
-                        ),
-                        file=sys.stderr
-                    )
+                    if not zk.check_node_existance("/consumers/{}".format(groupid)):
```
88manpreet: Can we try something like, 'if groupid in zk.get_children("/consumers")'?

dongw-yelp:
Took the comments.

dongw-yelp:
Pick up comments, compacted and simplified the logic.

88manpreet:
+1

PR #165, opened 2018-05-22
Delete the parameter "-e" "--exact" and add "-r" "--regex" for get_to…

YZG112358:
…pic_watermark. Set exactly topic matching as default to make the command more intuitively.

### DIFF
--- a/tests/acceptance/steps/util.py
+++ b/tests/acceptance/steps/util.py
@@ -189,11 +189,13 @@ def set_consumer_group_offset(topic, group, offset, storage='kafka'):
     )
 
 
-def call_watermark_get(topic_name, storage=None):
+def call_watermark_get(topic_name, storage=None, regex=False):
     cmd = ['kafka-consumer-manager',
            '--cluster-type', 'test',
            '--discovery-base-path', 'tests/acceptance/config',
            'get_topic_watermark', topic_name]
+    if regex:
+        cmd.extend(['-r'])
     return call_cmd(cmd)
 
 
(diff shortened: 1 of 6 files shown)


tests/acceptance/steps/watermark_get.py
```diff
@@ -31,3 +31,29 @@ def step_impl5(context):
     lowmark_pattern = 'Low Watermark: 0'
     assert highmark_pattern in context.output
     assert lowmark_pattern in context.output
+
+
+@when(u'we call the watermark_get command with -r')
+def step_impl6(context):
+    context.output = call_watermark_get('ab', regex=True)
```
88manpreet: Can we change this to 'bc' instead?

88manpreet:
+1 besides the above comments, lgtm

tests/acceptance/steps/util.py
```diff
@@ -189,11 +189,13 @@ def set_consumer_group_offset(topic, group, offset, storage='kafka'):
...
 
 
-def call_watermark_get(topic_name, storage=None):
+def call_watermark_get(topic_name, storage=None, regex=False):
     cmd = ['kafka-consumer-manager',
            '--cluster-type', 'test',
            '--discovery-base-path', 'tests/acceptance/config',
            'get_topic_watermark', topic_name]
+    if regex:
+        cmd.extend(['-r'])
```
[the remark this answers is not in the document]
YZG112358: Because other commands e.g 'call_offset_get' also uses extend. I try to keep the same style.

tests/acceptance/steps/watermark_get.py
billyevans: I would recommend to extend the topic names to something like:
abcd
abcde

To avoid suffix and prefix match with 'bc' search

YZG112358: Sure! 'bc' makes more sense. It should generate same output.

PR #166, opened 2018-05-23
Improve performance of broker decommission process

flavray:
## Description

(I'm currently only focusing on the decommissioning process. In order to keep the review process smooth, I'll start adding changes for rebalance and other commands once this PR is merged.)

    Generating a `_State` is an intensive task. Instead of generating a new
    `_State` after each new replica assignement, use the one that just got
    applied.
 

    Generating a full cluster assignment for a `_State` is an intensive
    task. Instead, only generate an assignment containing incremental
    changes incurred by the state changes.
 

    Finding the index of an element in a tuple multiple times is an
    intensive task. Instead, generate index reverse lookup maps once and use
    it each time.
 

    Make the decommission process repeatable by adding partitions in
    lexicographic order. This makes the verification easier. :-)

## Testing

* Basic unit tests + existing test suite are all green.
* Manually tested on our dev cluster. Output between the status quo and the new branch are the same.

## Performance

First results seems to indicate a 500x speed-up on the CPU-intensive part (excluding ZK+Metrics computation), from 125 minutes to 15 seconds 😄  (Using cProfile, which itself might have added a slight overhead)

(master)
screen shot 2018-05-23 at 3 42 04 pm

(u/flavr/KAFKA-9310-faster-decomission)
screen shot 2018-05-23 at 3 42 15 pm

Commands used to run the test: (spoiler alert: they are the same 😉)

(master)
```
time sudo -u root /opt/venvs/yelp-kafka-tool/bin/python -m cProfile -o output.prof /usr/bin/kafka-cluster-manager --cluster-type scribe     --cluster-name uswest1-devc     --logconf /nail/etc/kafka/conditional_rebalance_log.ini     --genetic-balancer     --partition-measurer yelp_kafka_tool.partition_measurement.pickled_signalfx_partition_measurer     --measurer-args ' --bytes-in-weight 0.25 --bytes-out-weight 0.25 --messages-in-weight 0.25 --partition-size-weight 0.25'     decommission 170396394     --max-partition-movements 250     --max-leader-changes 250
```

(u/flavr/KAFKA-9310-faster-decomission)
```
time sudo -u root /opt/venvs/yelp-kafka-tool/bin/python -m cProfile -o output.prof /usr/bin/kafka-cluster-manager --cluster-type scribe     --cluster-name uswest1-devc     --logconf /nail/etc/kafka/conditional_rebalance_log.ini     --genetic-balancer     --partition-measurer yelp_kafka_tool.partition_measurement.pickled_signalfx_partition_measurer     --measurer-args ' --bytes-in-weight 0.25 --bytes-out-weight 0.25 --messages-in-weight 0.25 --partition-size-weight 0.25'     decommission 170396394     --max-partition-movements 250     --max-leader-changes 250
```

### DIFF
--- a/kafka_utils/kafka_cluster_manager/cluster_info/genetic_balancer.py
+++ b/kafka_utils/kafka_cluster_manager/cluster_info/genetic_balancer.py
@@ -303,9 +303,9 @@ class GeneticBalancer(ClusterBalancer):
         # Add partition replicas to active brokers one-by-one.
         for partition_name in sorted(six.iterkeys(partitions)):  # repeatability
             partition = self.cluster_topology.partitions[partition_name]
-            count = partitions[partition_name]
+            replica_count = partitions[partition_name]
             try:
-                self.add_replica(partition_name, count)
+                self.add_replica(partition_name, replica_count)
             except InvalidReplicationFactorError:
                 raise BrokerDecommissionError(
                     "Not enough active brokers in the cluster. "
@@ -313,7 +313,7 @@ class GeneticBalancer(ClusterBalancer):
                     "but only {brokers} active brokers remain."
                     .format(
                         partition=partition_name,
-                        rf=partition.replication_factor + count,
+                        rf=partition.replication_factor + replica_count,
                         brokers=len(active_brokers)
                     )
                 )
@@ -346,8 +346,7 @@ class GeneticBalancer(ClusterBalancer):
                 )
             )
 
-        state = self.state
-        partition_index = state.partition_indices[partition]
+        partition_index = self.state.partition_indices[partition]
 
         for _ in range(count):
             # Find eligible replication-groups.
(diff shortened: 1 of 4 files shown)


kafka_utils/kafka_cluster_manager/cluster_info/genetic_balancer.py
```diff
@@ -369,7 +383,11 @@ def add_replica(self, partition_name, count=1):
 
             # Update cluster topology with highest scoring state.
             state = sorted(new_states, key=self._score, reverse=True)[0]
-            self.cluster_topology.update_cluster_topology(state.assignment)
+            self.cluster_topology.update_cluster_topology(state.pending_assignment)
+
+            # Update the internal state to match.
+            self.state = state
```
fede1024: Could we update and use `self.state` directly?

kafka_utils/kafka_cluster_manager/cluster_info/genetic_balancer.py
```diff
@@ -745,6 +767,10 @@ def __init__(self, cluster_topology, brokers=None):
             for rg in self.rgs
         )
 
+        # A tuple containing indices of partitions affect by changes made to
```
fede1024: nit: "affected"?

PR #168, opened 2018-05-29
Add NoNodeError to get_brokers & get_topics

YZG112358:
Catch Exceptions for commands under empty cluster exception

### DIFF
--- a/kafka_utils/util/zookeeper.py
+++ b/kafka_utils/util/zookeeper.py
@@ -101,8 +101,13 @@ class ZK:
 
         :rtype : dict of brokers
         """
-        broker_ids = self.get_children("/brokers/ids")
-
+        try:
+            broker_ids = self.get_children("/brokers/ids")
+        except NoNodeError:
+            _log.error(
+                "cluster is empty."
+            )
+            return {}
         # Return broker-ids only
         if names_only:
             return {int(b_id): None for b_id in broker_ids}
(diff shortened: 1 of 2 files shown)


kafka_utils/util/zookeeper.py
```diff
@@ -101,8 +101,13 @@ def get_brokers(self, names_only=False):
 
         :rtype : dict of brokers
         """
-        broker_ids = self.get_children("/brokers/ids")
-
+        try:
+            broker_ids = self.get_children("/brokers/ids")
+        except NoNodeError:
+            _log.error(
```
billyevans: Let' print a debug or info message here rather than error. And the same for get_topics()

tests/util/zookeeper_test.py
```diff
@@ -338,3 +338,17 @@ def test_get_topics(self, mock_client):
                 },
             }
             assert actual_without_fetch_state == expected_without_fetch_state
+
+            zk.get_children = mock.Mock(side_effect=NoNodeError())
```
billyevans: Why it's not new function ?

tests/util/zookeeper_test.py
```diff
@@ -338,3 +338,17 @@ def test_get_topics(self, mock_client):
...
+            actual_with_no_node_error = zk.get_topics()
+            expected_with_no_node_error = {}
+            zk.get_children.assert_called_with("/brokers/topics")
+            assert actual_with_no_node_error == expected_with_no_node_error
+
+    def test_get_brokers(self, mock_client):
```
billyevans: I think it's good idea to cover happy-case as well for get_brokers if it's not covered.

PR #169, opened 2018-06-04
kafka-check command checks for empty broker ids before running command

morningsend:
- refactored is_first_broker method 
- added unit tests

### DIFF
--- a/kafka_utils/kafka_check/commands/command.py
+++ b/kafka_utils/kafka_check/commands/command.py
@@ -49,6 +49,8 @@ class KafkaCheckCmd(object):
         self.cluster_config = cluster_config
         self.args = args
         with ZK(self.cluster_config) as self.zk:
+            broker_ids = get_broker_ids(self.zk)
+
             if args.controller_only and not is_controller(self.zk, args.broker_id):
                 terminate(
                     status_code.OK,
@@ -58,7 +60,15 @@ class KafkaCheckCmd(object):
                     ),
                     args.json,
                 )
-            if args.first_broker_only and not is_first_broker(self.zk, args.broker_id):
+            if args.first_broker_only and not broker_ids:
+                terminate(
+                    status_code.OK,
+                    prepare_terminate_message(
+                        'No brokers detected, nothing to check'
+                    ),
+                    args.json,
+                )
+            if args.first_broker_only and not is_first_broker(broker_ids, args.broker_id):
                 terminate(
                     status_code.OK,
                     prepare_terminate_message(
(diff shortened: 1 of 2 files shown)


kafka_utils/kafka_check/commands/command.py
```diff
@@ -80,6 +90,11 @@ def is_controller(zk, broker_id):
     return broker_id == zk.get_json('/controller').get('brokerid')
 
 
-def is_first_broker(zk, broker_id):
+def get_broker_ids(zk):
+    """Returns a list of integers for broker ids"""
+    return zk.get_brokers().keys()
```
flavray: Let's surround this with a `list`, to make sure it is compatible for both Python 2 and 3 🙂

morningsend: Thanks. I am fixing it.

kafka_utils/kafka_check/commands/command.py
```diff
@@ -80,6 +90,11 @@ def is_controller(zk, broker_id):
...
+
+
+def is_first_broker(broker_ids, broker_id):
     """Returns true if broker_id is the lowest broker id in the cluster, false otherwise."""
-    return broker_id == min(zk.get_brokers().keys())
+    return True if broker_ids and broker_id == min(broker_ids) else False
```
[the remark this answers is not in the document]
flavray: Sounds good, feel free to disregard the comment 🙂


Also, do you think we could make the boolean logic more readable by removing the `True if ... else False` part?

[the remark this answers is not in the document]
morningsend: I agree the condition should be more readable. Fixing :)

kafka_utils/kafka_check/commands/command.py
```diff
@@ -80,6 +90,14 @@ def is_controller(zk, broker_id):
...
-def is_first_broker(zk, broker_id):
+def get_broker_ids(zk):
+    """Returns a list of integers for broker ids"""
+    return list(zk.get_brokers())
+
+
+def is_first_broker(broker_ids, broker_id):
     """Returns true if broker_id is the lowest broker id in the cluster, false otherwise."""
-    return broker_id == min(zk.get_brokers().keys())
+    if not broker_ids:
```
flavray: This can probably just be `broker_ids and broker_id == min(broker_ids)` 🙂

morningsend: Python is a bit weird. I tried `[] and 4 == min([])` and got empty list `[]` rather than `False` Python's logical operators seem to do something inconsistent. We have to around it using `if` statement.

flavray: Oh. yeah. Python.

`len(broker_ids) > 0 and broker_id == min(zk.get_brokers().keys())` maybe? 😄

PR #170, opened 2018-06-06
--broker-ids option for rolling restart command

morningsend:
Added new optional argument `--broker-ids` to `kafka-rolling-restart` command:

    `kafka-rolling-restart --cluster-type xxxx --broker-ids x y z`

Will only restart brokers x, y, z in the cluster.
The new option accepts a space separated list of integer broker ids and will only restart the brokers specified by `--broker-ids` if it is present.

Also corrected documentation for `get_broker_list` function.

### DIFF
--- a/kafka_utils/kafka_rolling_restart/main.py
+++ b/kafka_utils/kafka_rolling_restart/main.py
@@ -74,8 +74,9 @@ def parse_opts():
         '--broker-ids',
         '-b',
         required=False,
-        type=str,
-        help='comma separated broker IDs to restart (optional, will restart all Kafka brokers in cluster if not specified)',
+        type=int,
+        nargs='+',
+        help='space separated broker IDs to restart (optional, will restart all Kafka brokers in cluster if not specified)',
     )
     parser.add_argument(
         '--discovery-base-path',
@@ -166,7 +167,7 @@ def parse_opts():
 
 
 def get_broker_list(cluster_config):
-    """Returns a dictionary of brokers in the form [(id: host)]
+    """Returns a list of brokers in the form [(id: host)]
 
     :param cluster_config: the configuration of the cluster
     :type cluster_config: map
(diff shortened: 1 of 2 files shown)


kafka_utils/kafka_rolling_restart/main.py
```diff
@@ -169,6 +176,22 @@ def get_broker_list(cluster_config):
         return [(id, data['host']) for id, data in brokers]
 
 
+def filter_broker_list(brokers, filter_by):
+    """Returns sorted list, a subset of elements from brokers in the form [(id, host)].
+    Passing empty list for filter_by will return all brokers.
```
flavray: nit: This function would probably be easier to read if this was not applied. Filtering by an empty list should return an empty list, and checks should be done before hand 🙂

I see that you're calling this function only if a list of brokers has been given as input, so it's definitely fine to skip this condition

kafka_utils/kafka_rolling_restart/main.py
```diff
@@ -70,6 +70,13 @@ def parse_opts():
         '-c',
         help='cluster name, e.g. "uswest1-devc" (defaults to local cluster)',
     )
+    parser.add_argument(
+        '--broker-ids',
+        '-b',
+        required=False,
+        type=str,
```
flavray: Have you considered using `argparse`'s `nargs` mechanism to directly have a list of integers here?
Using a comma-separated list of brokers is fine, but using built-in tooling could be a good idea

Context: https://docs.python.org/2/library/argparse.html#example

morningsend: sure I will take a look.

kafka_utils/kafka_rolling_restart/main.py
```diff
@@ -70,6 +70,14 @@ def parse_opts():
         '-c',
         help='cluster name, e.g. "uswest1-devc" (defaults to local cluster)',
     )
+    parser.add_argument(
+        '--broker-ids',
+        '-b',
+        required=False,
+        type=int,
+        nargs='+',
+        help='space separated broker IDs to restart (optional, will restart all Kafka brokers in cluster if not specified)',
```
88manpreet: Looks great. Minor comment for doc change.
I think it will restart all brokers based on whether -n is provided or not.

PR #174, opened 2018-06-19
Add brokers as follower and leader when decommissioning brokers.

flavray:
The best of the two will be picked.

When replacing a broker with a new one, it often happens that the new
broker is empty and can be picked as leader for all/most of the
partitions on the broker to be decommissioned.
This change will explore adding each broker to the replica set of
partitions as a follower, and as a leader.

### DIFF
--- a/kafka_utils/kafka_cluster_manager/cluster_info/genetic_balancer.py
+++ b/kafka_utils/kafka_cluster_manager/cluster_info/genetic_balancer.py
@@ -370,15 +370,21 @@ class GeneticBalancer(ClusterBalancer):
                 if rg.count_replica(partition) < opt_replicas
             ] or non_full_rgs
 
-            # Add the replica to every eligible broker.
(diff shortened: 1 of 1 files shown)


yemyatthein:
I presume the change is intended for decommissioning where the terminating broker may have leader partitions and we want to have the same setting (same leaders) in a new cluster state by giving an equal chance (being leader or follower) to the scoring function. I was wondering if this would unnecessarily trigger leader elections if the use case is to increase the replica of a partition.

Other than that, the change itself looks good to me.

flavray:
It will definitely trigger some leader elections: partitions for whom the decommissioned was the leader will probably end up having the newly commissioned broker to be their leader (instead of just being a follower).

This is the behaviour we'd want. History seems to show that new brokers are leaders for few partitions, adding load on other brokers. Hopefully this change will move leadership to the new broker.

Partitions that still have their leader won't be affected.

PR #183, opened 2018-08-07
Show error message for nonexistent group in offset_get

Baisang:
We generate the list of groups in `list_groups` command by calling `kafka_group_reader.read_groups().keys()`; so if a group is not in this list, it doesn't exist.

`kafka_group_reader.read_group()` is just a wrapper around the `read_groups()` method that gets the specific `group_id` entry from all of the groups (i.e., `read_groups().get(group_id, [])`. So if there are no topics for the group (`group_topics == []`) the group does not exist

### DIFF
--- a/kafka_utils/kafka_consumer_manager/commands/offset_manager.py
+++ b/kafka_utils/kafka_consumer_manager/commands/offset_manager.py
@@ -131,7 +131,15 @@ class OffsetManagerBase(object):
             groupid
     ):
         kafka_group_reader = KafkaGroupReader(cluster_config)
-        return kafka_group_reader.read_group(groupid)
+        group_topics = kafka_group_reader.read_group(groupid)
+        if len(group_topics) == 0:
+            print(
+                "Error: Either Consumer Group {groupid} has never committed offsets, or the group does not exist.".format(
+                    groupid=groupid,
+                ),
+                file=sys.stderr,
+            )
+        return group_topics
 
     @classmethod
     def get_topics_for_group_from_zookeeper(


kafka_utils/kafka_consumer_manager/commands/offset_manager.py
```diff
@@ -131,7 +131,13 @@ def get_topics_for_group_from_kafka(
...
     ):
         kafka_group_reader = KafkaGroupReader(cluster_config)
-        return kafka_group_reader.read_group(groupid)
+        group_topics = kafka_group_reader.read_group(groupid)
+        if len(group_topics) == 0:
+            print(
+                "Error: Consumer Group ID {groupid} does not exist.".format(groupid=groupid),
+                file=sys.stderr,
+            )
+        return group_topics
```
ashsingh21: why are we returning empty group_topics?

Baisang: Wanted to keep current behavior where `get_topics_for_group_from_kafka()` does not `fail_on_error`. But if we think it should, then I can change that in a separate commit (I think it makes more sense to `fail_on_error`, i.e. make failing behavior match when we query Zookeeper by having it `sys.exit(1)` etc. But I'm not sure if this could break other things).

88manpreet: The consumer-group can not exist for two reasons:
a) The consumer actually does not exist.
b) The consumer exists and is running but has never committed offsets. As a result kafka cannot find it.
We should modify the log accordingly to avoid confusion for the team.
This can be tested and confirmed by running a new-consumer and having offsets committed (auto_offset_reset=False and no manual commit). Ref.: https://yelpwiki.yelpcorp.com/display/DISTSYS/Yelp+Kafka

Also, I think it's offset_set command with --force option, where would want to return the empty result without exiting. The rationale behind this is that we can still commit offsets (set-offsets) even if consumer-group is not found. This is worth testing with above setup too.

Baisang: Tested this behavior with a simple consumer.

b) A consumer that consumes without committing offsets (`enable_auto_commit=False`) cannot be found by kafka
c) Using `offset_set --force` when the consumer does not exist in Kafka commits the offset correctly without an error.

ashsingh21: I see "brod" is using that method, might break it if we do sys.exit

kafka_utils/kafka_consumer_manager/commands/offset_manager.py
```diff
@@ -131,7 +131,15 @@ def get_topics_for_group_from_kafka(
...
-        return kafka_group_reader.read_group(groupid)
+        group_topics = kafka_group_reader.read_group(groupid)
+        if len(group_topics) == 0:
+            print(
+                "Error: Either Consumer Group {groupid} has never committed offsets, or the group does not exist.".format(
+                    groupid=groupid,
+                ),
+                file=sys.stderr,
+            )
+        return group_topics
```
88manpreet: +1

PR #189, opened 2018-09-07
Fix bug in unsubscribe_topics with --topics flag

Baisang:
This fixes a bug where using unsubscribe_topics with the --topics flag
displayed all of the subscribed topics of the given consumer group as
subject to change when only the topics supplied with --topics should be
changed.

The bug was cosmetic only, so no actual functionality was changed.

### DIFF
--- a/kafka_utils/kafka_consumer_manager/commands/offset_manager.py
+++ b/kafka_utils/kafka_consumer_manager/commands/offset_manager.py
@@ -50,11 +50,26 @@ class OffsetManagerBase(object):
         client,
         storage='kafka',
         fail_on_error=True,
-        quiet=False
+        quiet=False,
+        topics=None,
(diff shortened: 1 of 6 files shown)


kafka_utils/kafka_consumer_manager/commands/offset_manager.py
```diff
@@ -206,6 +207,13 @@ def preprocess_args(
                 client,
             )
 
+        # If a list of topics is supplied, let's filter out the subscribed
```
88manpreet: Looks good. Would it be confusing if both topic and topics is provided?

88manpreet:
+1 once all the tests passes

88manpreet:
+1 once all the tests passes

PR #192, opened 2018-10-04
Safe zero offsets in offset_set

Baisang:
Previously, using offset_set was dangerous if any partition was
specified to be set to offset zero. It would erase all offsets for the
entire topic, not just the topic partition.

This change addresses the issue by first saving the current offsets,
setting the zero offsets, and then setting the non-zero offsets
(including the current offsets) so that no offsets are lost.


Baisang:
Nevermind, it's a bug somewhere else in our client

PR #193, opened 2018-10-05
Fix bug in offset_get that did not display all offsets

Baisang:
We noticed a bug in offset_get where we did not display offsets for a
topic if a consumer group committed a zero-offset for any partition of
the topic.

This patch ensures that we only hide offsets for a topic if a consumer
group has committed a zero-offset for every partition of the topic
(hence, unsubscribing)


88manpreet:
+1

88manpreet:
+1

PR #194, opened 2018-10-09
Allow for yelp-only usage of internal pypi

Baisang:
This is based off of https://github.com/Yelp/paasta/commit/49c1cd0c9c8c641ad3332360ba3c09ad9d8cac4c

Basically, we won't have to do the export PIP_INDEX_URL=... internally anymore.

I tested this both internally and externally, and it seemed to work.


Makefile
```diff
@@ -1,3 +1,12 @@
+# Set ENV to 'YELP' if FQDN ends in '.yelpcorp.com'
```
flavray: nit: I don’t think this comment applies here 🙂

Baisang: You're right, I misread what it meant haha

PR #200, opened 2018-12-03
Change import path for error classes #199

### LINKED ISSUE #199: ImportErrors - Structure of dependent kafka-python package changed
The code-structure of the python-kafka package has changed which is causing some parts of this package to import classes with the wrong path. This can be seen in [util/offsets.py](https://github.com/Yelp/kafka-utils/blob/master/kafka_utils/util/offsets.py) file which imports error classes from _kafka.structs_. These are however not anymore in the structus.py file, but in the errors.py file (as can be seen in this [commit](https://github.com/dpkp/kafka-python/commit/221f56d8a05cdc2d37f85018e4af352b4b2a95c5#diff-d44163af6a47884d5684568142e14ac7R49))

suknuk:
Even though there is a _from kafka.errors import *_ at the end of the kafka/structs.py file to support legacy imports, the change to import from the actual class fixes my bug of unresolved imports. Fixes #199

### DIFF
--- a/kafka_utils/util/offsets.py
+++ b/kafka_utils/util/offsets.py
@@ -19,8 +19,8 @@ from collections import namedtuple
 
 import six
 from kafka.common import UnknownTopicOrPartitionError
-from kafka.structs import BrokerResponseError
-from kafka.structs import check_error
+from kafka.errors import BrokerResponseError
+from kafka.errors import check_error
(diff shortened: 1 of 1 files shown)


88manpreet:
Thanks @suknuk for noticing this. We included the fix as part of our latest version.

PR #215, opened 2019-03-27
Add support for --max-movement-size argument in decommission

Baisang:
I still need to test this internally, but want to get the code checked out first.

This will allow users to limit the size of decommission movements not just by # of partitions, but also by data size. The goal is to give users more flexibility in how they want to do decommissions.

By default, there is no data size limit, which makes this backwards compatible.

### DIFF
--- a/kafka_utils/kafka_cluster_manager/cmds/command.py
+++ b/kafka_utils/kafka_cluster_manager/cmds/command.py
@@ -28,6 +28,9 @@ from kafka_utils.util.validation import assignment_to_plan
 from kafka_utils.util.zookeeper import ZK
 
 
+DEFAULT_MAX_MOVEMENT_SIZE = float('inf')
+
+
 class ClusterManagerCmd(object):
     """Interface used by all kafka_cluster_manager commands
     The attributes cluster_config, args and zk are initialized on run().
@@ -157,9 +160,10 @@ class ClusterManagerCmd(object):
     def get_reduced_assignment(
         self,
         original_assignment,
-        new_assignment,
+        cluster_topology,
         max_partition_movements,
         max_leader_only_changes,
+        max_movement_size=DEFAULT_MAX_MOVEMENT_SIZE,
     ):
         """Reduce the assignment based on the total actions.
 
@@ -170,15 +174,18 @@ class ClusterManagerCmd(object):
 
         Argument(s):
         original_assignment:    Current assignment of cluster in zookeeper
-        new_assignment:         New proposed-assignment of cluster
+        cluster_topology:       Cluster topology containing the new proposed-assignment of cluster
         max_partition_movements:Maximum number of partition-movements in
                                 final set of actions
         max_leader_only_changes:Maximum number of actions with leader only changes
+        max_movement_size:      Maximum size, in bytes, to move in final set of actions
         :return:
         :reduced_assignment:    Final reduced assignment
         """
+        new_assignment = cluster_topology.assignment
         if (not original_assignment or not new_assignment or
-                max_partition_movements < 0 or max_leader_only_changes < 0):
+                max_partition_movements < 0 or max_leader_only_changes < 0 or
+                max_movement_size < 0):
             return {}
 
         # The replica set stays the same for leaders only changes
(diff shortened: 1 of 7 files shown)


kafka_utils/kafka_cluster_manager/cmds/command.py
```diff
@@ -157,9 +157,10 @@ def process_assignment(self, assignment, allow_rf_change=False):
     def get_reduced_assignment(
         self,
         original_assignment,
-        new_assignment,
+        cluster_topology,
         max_partition_movements,
         max_leader_only_changes,
+        max_movement_size=float('inf'),
```
88manpreet: Minor: Should we just import and use DEFAULT_MAX_MOVEMENT_SIZE here for consistency?

kafka_utils/kafka_cluster_manager/cmds/command.py
```diff
@@ -248,19 +257,22 @@ def _extract_actions_unique_topics(self, movement_counts, max_movements):
         # Create reduced assignment minimizing duplication of topics
         extracted_actions = []
         curr_movements = 0
+        curr_size = 0
         action_available = True
-        while curr_movements < max_movements and action_available:
+        while curr_movements < max_movements and curr_size < max_movement_size and action_available:
```
88manpreet: To be on safer side, should we at least allow 1 partition for movement if all the partitions exceed the given size.
Another alternate is to through a warning/error log if such case arises.

Baisang: Yeah, this is a little tricky since we want to have a good value. The issue arises when clusters have high variance in partition size. In these situations, if we go by the largest partition in the cluster, we won't be able to properly utilize the `max_movement_size` well. But if we don't we'll get to a state where we cannot make progress.

I think adding an `auto-max-movement-size` like in rebalance and setting that to the size of the largest partition, and then printing a warning if the user specified a `max-movement-size` that is less than the size of the largest partition would be enough.

Baisang:
I tested manually and both options seem to work as expected in the generated plan.

kafka_utils/kafka_cluster_manager/cmds/command.py
ecanzonieri: Rebalance exits 1 when the max movement size is lower than the biggest partition in the cluster. I think we should do the same in decommission. For decommission failing fast when the process cannot be completed is even more important than rebalance. Perhaps as an optimization we can fail if the max-movement-size is smaller than the largest partition to decommssion.

Baisang: ++ That makes a lot of sense. Having the optimization seems doable as well, as we can just iterate through the partitions of the brokers that are to be decommissioned.

Baisang:
I made the changes @ecanzonieri suggested. Things seem to work in my manual testing.

88manpreet:
+1 once the tests pass.

Baisang:
Looking into the test failures. Looks like list_groups is failing, fails on master branch too so I don't think its related to this change. Will look into it anyway.

Baisang:
```Feature: kafka_consumer_manager list_groups subcommand # tests/acceptance/list_groups.feature:1

  Scenario: Calling the list_group command          # tests/acceptance/list_groups.feature:3
    Given we have initialized kafka offsets storage # tests/acceptance/steps/common.py:60
    Given we have initialized kafka offsets storage # tests/acceptance/steps/common.py:60 25.277s
    Given we have a set of existing consumer groups # tests/acceptance/steps/list_groups.py:28
    Given we have a set of existing consumer groups # tests/acceptance/steps/list_groups.py:28 2.643s
    When we call the list_groups command            # tests/acceptance/steps/list_groups.py:49 0.261s
    Then the groups will be listed                  # tests/acceptance/steps/list_groups.py:54
> /work/tests/acceptance/steps/list_groups.py(58)step_impl5()
-> for group in context.groups:
(Pdb) context.output
'Traceback (most recent call last):\n  File "/work/.tox/py27-acceptance/bin/kafka-consumer-manager", line 6, in <module>\n    run()\n  File "/work/.tox/py27-acceptance/local/lib/python2.7/site-packages/kafka_utils/kafka_consumer_manager/main.py", line 121, in run\n    arg
s.command(args, conf)\n  File "/work/.tox/py27-acceptance/local/lib/python2.7/site-packages/kafka_utils/kafka_consumer_manager/commands/list_groups.py", line 56, in run\n    kafka_groups = cls.get_kafka_groups(cluster_config)\n  File "/work/.tox/py27-acceptance/local/lib/
python2.7/site-packages/kafka_utils/kafka_consumer_manager/commands/list_groups.py", line 37, in get_kafka_groups\n    return list(kafka_group_reader.read_groups().keys())\n  File "/work/.tox/py27-acceptance/local/lib/python2.7/site-packages/kafka_utils/kafka_consumer_man
ager/util.py", line 290, in read_groups\n    for p in self.consumer.partitions_for_topic(CONSUMER_OFFSET_TOPIC)\nTypeError: \'NoneType\' object is not iterable\n'```

Traceback seems to indicate for some reason partitions_for_topic is returning None

Baisang:
Fixed the bug in https://github.com/Yelp/kafka-utils/pull/216, merged that to master and rebased this branch off of master.

PR #216, opened 2019-04-02
Fetch metadata for KafkaGroupReader consumer before partitions_for_topic

Baisang:
Applies the fix observed in https://github.com/dpkp/kafka-python/issues/1742 to fix integration tests


88manpreet:
+1 once the build passes

PR #217, opened 2019-04-03
Upgrade requirements

Baisang:
Upgrade a lot of our requirements to newer/later versions. This should address some of the CVEs in issues.

make test and make acceptance both passed


88manpreet:
+1

PR #218, opened 2019-04-05
Fix bug if --max-movement-size=0

Baisang:
There's a bug in decommission if max_movement_size becomes 0 and there are only 0-sized partitions left. This will mean no progress will be made.

This also makes the script more efficient in adding as many 0-size partitions as possible (up to given leader/partition movement cap) when doing other decommissions.


88manpreet:
+1

PR #219, opened 2019-04-06
max movement size improvements

Baisang:
Need to do some more manual testing still.

- I noticed decommission still uses the scoring in genetic balancer, which takes in `self.args.max_movement_size` (passed in from the cmd line). In the previous change to decommission where we added this argument, we set the default value to be `float('inf')`. The issue is that now this `float('inf')` will be used in genetic balancer for scoring even if the user did not pass in `--max-movement-size`. So we set the default value to be `None` in the parser (matching the rebalance command) and only set it to `float('inf')` when we reduce the overall assignment.
- I also noticed that passing in `--max-movement-size 0` will cause divide-by-zero error in genetic balancer, so I just coerced it to 1 to avoid the issue in the balancer. Same with `--max-leader-changes`
- Lastly, a new `--force-progress` flag which, when combined with `--max-movement-size`, will ensure that progress will be made during decommission in case `--max-movement-size` is less than one of the partitions left to move. Basically, if during the reducing assignment phase we have reduced the assignment to 0 actions, we recompute reduction using a larger `--max-movement-size` (set to the largest partition left to move). It should only kick in at the tail end of a decommission.

### DIFF
--- a/kafka_utils/kafka_cluster_manager/cluster_info/genetic_balancer.py
+++ b/kafka_utils/kafka_cluster_manager/cluster_info/genetic_balancer.py
@@ -607,13 +607,17 @@ class GeneticBalancer(ClusterBalancer):
             max_score += self.args.broker_leader_count_score_weight
 
         if self.args.max_movement_size is not None and score_movement:
+            # Avoid potential divide-by-zero error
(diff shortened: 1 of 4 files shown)


88manpreet:
+1

PR #220, opened 2019-04-13
Speed up rebalance plan execution

udaysagar2177:
This speeds up the plan execution step as it will fetch the current plan only for the topics in proposed plan.


udaysagar2177:
@flavray Let me address your concerns by this weekend, thanks for the review!

udaysagar2177:
@flavray I addressed your comments. The build failed this time at `docker-compose build` step, before executing any tests and the error appears to be a transient one. How can I re-trigger the build?

flavray:
I've kicked off another build, let's see how it goes 🙂

udaysagar2177:
My pleasure, thanks for the quick review!

PR #221, opened 2019-04-25
Implement --head option for kafka-check command

billyevans:
Add few more tests for formatting functions of subcommands of kafka-check

We need that param to be able to pass affected topics/partitions into sensu.
Bash `| head/tail` pipeline doesn't work properly because of unusual handling of SIGPIPE on python side. Also bash pipe hides exit code of the first command. Python solution seems more sustainable.

### DIFF
--- a/kafka_utils/kafka_check/commands/offline.py
+++ b/kafka_utils/kafka_check/commands/offline.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 from __future__ import absolute_import
 
+import itertools
 import sys
 
 from kafka_utils.kafka_check import status_code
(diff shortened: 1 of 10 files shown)


kafka_utils/kafka_check/commands/offline.py
```diff
@@ -42,18 +43,21 @@ def run_command(self):
         )
 
         errcode = status_code.OK if not offline else status_code.CRITICAL
-        out = _prepare_output(offline, self.args.verbose)
+        out = _prepare_output(offline, self.args.verbose, self.args.head)
         return errcode, out
 
 
-def _prepare_output(partitions, verbose):
+def _prepare_output(partitions, verbose, head_limit):
```
yzernik: Would it make sense to have some kind of generic prepare_output function that can be re-used? Would that add too much complication?

billyevans: I was thinking about it multiple times before. Have a look at _prepare_output() - 80-90% of it is transformation from internal-representation to string. That transformation is very command-specific, so it won't go into shared place. Having generic function will make everything way harder to read, probably it will safe 10 lines by price of simplicity. Maybe with some tricks we can make it nicer. However, it's command line tools, I don't we should be perfect here.

yzernik: Sounds good.

yzernik:
+1

PR #223, opened 2019-04-29
Relax requirements for pytz and pyyaml

Baisang:
Included PyYAML here since we switched to 5.10 in requirements.txt and didn't notice anything odd.


billyevans:
Looks good for me!

PR #225, opened 2019-06-10
Human friendly output for bytes in decommission

Baisang:
Adds a new dependency humanfriendly that we can start using to display bytes and other numbers in human friendly values

Some sample output

```
WARNING:DecommissionCmd:Max partition movement size is only 13.0 bytes, but remaining partitions to move range from 0.0 bytes to 4.12 GB. The decommission may be slower than expected
INFO:DecommissionCmd:Auto-max-movement-size: using 4.12 GB (4120645437.0) as max-movement-size.
ERROR:DecommissionCmd:Max partition movement size is only 1.0 byte, but remaining partitions to move range from 0.0 bytes to 4.12 GB. The decommission will not make progress
```

Wasn't able to test the case in `command.py` but it should work the same.

### DIFF
--- a/kafka_utils/kafka_cluster_manager/cmds/command.py
+++ b/kafka_utils/kafka_cluster_manager/cmds/command.py
@@ -19,6 +19,7 @@ import logging
 import sys
 from collections import defaultdict
 
+import humanfriendly
 import six
 from six.moves import input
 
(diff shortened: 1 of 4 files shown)


billyevans:
Looks good to me!

88manpreet:
+1

PR #230, opened 2019-07-18
Fix configuration writing in kafka-manual-throttle

flavray:
Instead of writing configuration in the `config` dict in Zookeeper,
kafka-manual-throttle was writing updates at the config toplevel. This
meant that Kafka was ignoring them.

Before:
{'config': {}, 'leader.replication.throttled.rate': XXXX}

After:
{'config': {'leader.replication.throttled.rate': XXXX}}

### DIFF
--- a/kafka_utils/kafka_manual_throttle/main.py
+++ b/kafka_utils/kafka_manual_throttle/main.py
@@ -157,7 +157,7 @@ def read_throttles(zk, brokers):
     throttles = {}
 
     for broker_id in brokers:
-        config = zk.get_broker_config(broker_id)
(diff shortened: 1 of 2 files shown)


flavray:
As far as I can tell, the top-level dictionary only contains `config` and `version`. I don't think it's got any other key.

flavray:
Ok, according to this [document|https://cwiki.apache.org/confluence/display/KAFKA/Kafka+data+structures+in+Zookeeper#KafkadatastructuresinZookeeper-0.9.0], these are indeed the only 2 keys in the top-level.
Not sure how up to date this document is, but I've been using it as my bible for years now 😄

PR #231, opened 2019-07-18
Add warning if timestamp argument in offsets_for_timestamp is in seconds

mhaseebmlk:
Print a warning message if the argument to the offsets_for_timestamp flag is not in milliseconds.

### DIFF
--- a/kafka_utils/kafka_consumer_manager/commands/offsets_for_timestamp.py
+++ b/kafka_utils/kafka_consumer_manager/commands/offsets_for_timestamp.py
@@ -83,7 +83,7 @@ class OffsetsForTimestamp(OffsetManagerBase):
 
     @classmethod
     def print_offsets(cls, partition_to_offset, orig_timestamp):
-        milliseconds_thresold = 999999999999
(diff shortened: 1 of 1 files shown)


kafka_utils/kafka_consumer_manager/commands/offsets_for_timestamp.py
```diff
@@ -83,6 +83,19 @@ def run(cls, args, cluster_config):
 
     @classmethod
     def print_offsets(cls, partition_to_offset, orig_timestamp):
+        milliseconds_thresold = 999999999999
```
Baisang: Can we add a comment here explaining why we have the threshold? It might not be clear to those reading

mhaseebmlk: Just added a comment. Can you please have a look? I feel like the comment is too long

Baisang: it's fine!

Issue #214, opened 2019-03-08
kafka-consumer-manager doesn't work with kafka 2.1


boniek83:
Probably due to group metadata changes in new kafka. Similar problem in burrow: https://github.com/linkedin/Burrow/issues/505

klass-ivan:
Any updates on this? I'm trying to use kafka-consumer-manager on 2.1 which result in ``` line 892, in check_version
    raise Errors.NoBrokersAvailable()```  but I'm not 100% sure if it's because of this issue or configuration problem

PR #236, opened 2019-09-05
Support consumer offset message value schema v3

mborst:
This should fix https://github.com/Yelp/kafka-utils/issues/214.


Baisang:
Thanks for contributing and sorry for the delay -- promise I'll get to this, just need to do some internal testing

jow-:
We successfully tested this change in our environment against Kafka 2.13-2.4.0 in your environment. 

Without this PR, we were unable to retrieve groups and offsets from our cluster.

Baisang:
@jow- Sorry about the delay. Unfortunately I don't get paid to work on open source all the time :)

I am testing these changes locally and will include them as part of new release 3.3.0 at the end of this week assuming things go well (don't see why not)

PR #237, opened 2019-09-17
KAFKA-23818_relax_pytz_version

yzernik:
This pinned version is too strict and it is causing dependency conflicts in some places.
Relaxing the requirement in setup.py looks like it should be safe.


88manpreet:
+1

PR #238, opened 2019-10-02
[kafka-consumer-manager] Not showing error for successful commit

transcedentalia:
Should fix KAFKA016447 :)

### DIFF
--- a/kafka_utils/kafka_consumer_manager/commands/offset_manager.py
+++ b/kafka_utils/kafka_consumer_manager/commands/offset_manager.py
@@ -144,13 +144,22 @@ class OffsetManagerBase(object):
         groupid,
     ):
         kafka_group_reader = KafkaGroupReader(cluster_config)
-        group_topics = kafka_group_reader.read_group(groupid)
-        if len(group_topics) == 0:
+        try:
+            group_topics = kafka_group_reader.read_group(groupid)
+        except KeyError:
             print(
-                "Error: Either Consumer Group {groupid} has never committed offsets, or the group does not exist.".format(
+                "Error: Consumer Group {groupid} does not exist.".format(
                     groupid=groupid,
                 ),
                 file=sys.stderr,
+            )
+            group_topics = []
+        if len(group_topics) == 0:
+            print(
+                "Info: It's the first time Consumer Group {groupid} is committing offsets.".format(
+                    groupid=groupid,
+                ),
+                file=sys.stdout,
             )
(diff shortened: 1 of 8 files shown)


kafka_utils/kafka_consumer_manager/commands/offset_manager.py
```diff
@@ -144,14 +144,23 @@ def get_topics_for_group_from_kafka(
...
     ):
         kafka_group_reader = KafkaGroupReader(cluster_config)
-        group_topics = kafka_group_reader.read_group(groupid)
-        if len(group_topics) == 0:
+        try:
+            group_topics = kafka_group_reader.read_group(groupid)
+        except KeyError:
             print(
-                "Error: Either Consumer Group {groupid} has never committed offsets, or the group does not exist.".format(
+                "Error: Consumer Group {groupid} does not exist.".format(
```
poros: What about changing it to something like "Warning: Consumer Group X does not exist. Creating it."? It is not really an error and we just go forward anyway.

Baisang: This function is called in other situations as well. I think that's why I made the print message intentionally vague at the time :/

kafka_utils/kafka_consumer_manager/commands/offset_manager.py
```diff
@@ -144,14 +144,23 @@ def get_topics_for_group_from_kafka(
...
                     groupid=groupid,
                 ),
                 file=sys.stderr,
             )
+            group_topics = []
+        if len(group_topics) == 0:
+            print(
```
Baisang: Wonder if this print can be put in a different function, that would make the function itself more reusable. Are there times we call this function when we are not committing offsets?

I think in general the functions in `offset_manager` were structured a little weird. There's a tight coupling of the print messages and the behavior

transcedentalia:
@poros The tests are failing on the master branch with the same errors and are unrelated to my change. Despite the errors, `make test` on the devbox is marked as successful. :shrug:

kafka_utils/kafka_consumer_manager/commands/offset_manager.py
transcedentalia: If I understand correctly it's always called when committing offsets.

transcedentalia:
@Baisang, @poros I've created KAFKA-24674 for the travis docker errors.

PR #242, opened 2019-11-05
Fix travis build by unpinning docker version to be installed

Baisang:
Looks like we were trying to install an older version of docker over a new one.

My guess is the new one had some confs for systemd and the old one had some confs for upstart, and things didn't get cleaned up properly.

Let's just use the newer one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,13 +6,6 @@ sudo: required
 services:
 - docker
 before_install:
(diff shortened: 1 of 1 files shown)


Baisang:
Something weird happened to Github, but the commit for this is in master so we can say it has been merged. Closing

PR #245, opened 2019-12-03
KAFKA-18193: catch UnknownMemberIdError when unsubscribing from topic…

lennrt:
…s and improve error message

### DIFF
--- a/kafka_utils/kafka_consumer_manager/commands/unsubscribe_topics.py
+++ b/kafka_utils/kafka_consumer_manager/commands/unsubscribe_topics.py
@@ -16,6 +16,7 @@ from __future__ import absolute_import
 from __future__ import print_function
 
 import six
+from kafka.errors import UnknownMemberIdError
 
 from .offset_manager import OffsetWriter
 from kafka_utils.util.client import KafkaToolClient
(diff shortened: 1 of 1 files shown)


kafka_utils/kafka_consumer_manager/commands/unsubscribe_topics.py
```diff
@@ -107,7 +108,17 @@ def unsubscribe_topics(self, group, topics, partitions, topics_dict):
             self.unsubscribe_partitions(group, topics[0], partitions)
         elif topics:
             for topic in topics:
-                self.delete_topic(group, topic)
+                try:
+                    self.delete_topic(group, topic)
+                except UnknownMemberIdError:
+                    print(
```
lsterk: I think we should print to `stderr` here, this seems consistent with other usage ([example](https://github.com/Yelp/kafka-utils/blob/4b2c04b329e5847019ea490ccd72d97812681c70/kafka_utils/kafka_consumer_manager/commands/offset_set_timestamp.py#L46))

>You must ensure none of the consumers with this consumer group id are running

Is it more clear to say "Please ensure all consumers in this consumer group are stopped" or similar? It makes more sense to me, but your call on this part.

PR #248, opened 2020-02-26
Add support for removing brokers in replace

Baisang:
This adds support for removing brokers from replica assignment. We specify a replica removal by having a source-broker with no dest-broker. Replica addition could also be added in a similar way, but I'm gonna leave that for the next PR.

Before it was not possible because the script would do two steps of verification for replacement:

1. Check that the overall replication factor of the topic partitions has not changed. This check can now be skipped with the `--rf-changes` flag
2. Check that the replication factor of topic partitions within the same topic is the same. This check can now be skipped with the `--rf-mismatch` flag.

We also introduce a `--topic-partition-filter` flag, that is a path to some output from `kafka-check`, so that you can focus on fixing offline or specific partitions instead of generating a plan for all of them.

Wrote some unit tests and will be trying this out in dev later today, just want to get the review out first.

A sample invocation to generate a reassignment plan to remove a faulty broker id `0` might be:

`kafka-cluster-manager -t test_cluster --write-to-file reassign_plan.json replace-broker --source-broker 0 --max-leader-changes 5000 --max-partition-movements 5000 --topic-partition-filter topic_partitions.txt --rf-mismatch  --rf-change`

where `topic_partitions.txt` is similar to `kafka-check` output:

```
my_topic:0
other_topic:1
```

While the above command is rather long, it should be much better than relying on adhoc scripts to generate reassignment json files.

### DIFF
--- a/kafka_utils/kafka_cluster_manager/cluster_info/cluster_topology.py
+++ b/kafka_utils/kafka_cluster_manager/cluster_info/cluster_topology.py
@@ -155,20 +155,24 @@ class ClusterTopology(object):
         """Move all partitions in source broker to destination broker.
 
         :param source_id: source broker-id
-        :param dest_id: destination broker-id
+        :param dest_id: destination broker-id, or None (to remove source broker)
         :raises: InvalidBrokerIdError, when either of given broker-ids is invalid.
         """
         try:
             source = self.brokers[source_id]
-            dest = self.brokers[dest_id]
+            dest = None
+            if dest_id is not None:
+                dest = self.brokers[dest_id]
             # Move all partitions from source to destination broker
             for partition in source.partitions.copy():  # Partitions set changes
                 # We cannot move partition directly since that re-orders the
                 # replicas for the partition
                 source.partitions.remove(partition)
-                dest.partitions.add(partition)
+                if dest:
+                    dest.partitions.add(partition)
                 # Replace broker in replica
                 partition.replace(source, dest)
+
         except KeyError as e:
             self.log.error("Invalid broker id %s.", e.args[0])
             raise InvalidBrokerIdError(
--- a/kafka_utils/kafka_cluster_manager/cmds/replace.py
+++ b/kafka_utils/kafka_cluster_manager/cmds/replace.py
@@ -18,6 +18,8 @@ from __future__ import print_function
 import logging
 import sys
 
+import six
+
 from .command import ClusterManagerCmd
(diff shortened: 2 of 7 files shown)


kafka_utils/kafka_cluster_manager/cluster_info/cluster_topology.py
```diff
@@ -155,20 +155,24 @@ def replace_broker(self, source_id, dest_id):
...
             # Move all partitions from source to destination broker
             for partition in source.partitions.copy():  # Partitions set changes
                 # We cannot move partition directly since that re-orders the
                 # replicas for the partition
                 source.partitions.remove(partition)
-                dest.partitions.add(partition)
+                if dest:
+                    dest.partitions.add(partition)
                 # Replace broker in replica
                 partition.replace(source, dest)
```
mhaseebmlk: Is this a no-op when `dest == None`?

mhaseebmlk: I see the logic for this was added in `partition.py`. Resolving comment.

kafka_utils/kafka_cluster_manager/cmds/replace.py
```diff
@@ -50,8 +52,8 @@ def build_subparser(self, subparsers):
         subparser.add_argument(
             '--dest-broker',
             type=int,
-            required=True,
-            help='Broker id of destination broker.',
+            default=None,
+            help='Broker id of destination broker. If empty, will shrink the replica set of topics',
```
mhaseebmlk: What will happen if this is called multiple times without a dest-broker such that the replicas becomes 0?

mhaseebmlk: nit: Although unlikely but what will happen if this script is run multiple times without a dest-broker such that the replicas for a topic become 0? Should we have a check that prevents that from happening?

Baisang: That is already checked, it'll fail in that case.

For example, if I do this on our test cluster (which only has 3 brokers, and some topics with only 1 replica), removing a broker will result in the following error:

```
ERROR:kafka_utils.util.validation:Non-empty "replicas" expected: {'topic': 'my_topic', 'partition': 4, 'replicas': []}
```

mhaseebmlk: Sounds good!

PR #252, opened 2020-03-09
Allow using Kafka AdminClient for kafka-consumer-manager commands

flavray:
Reading the __consumer_offsets topic at every run is slow and can add
unwanted load on a Kafka cluster. The AdminClient allows to get the list
of consumer groups and the list of topics from a group id in constance
time.

This is currently a feature flag to keep backwards compatibility, but
will later be enabled by default.

--use-admin-client

### DIFF
--- a/kafka_utils/kafka_consumer_manager/commands/list_groups.py
+++ b/kafka_utils/kafka_consumer_manager/commands/list_groups.py
@@ -55,7 +55,8 @@ class ListGroups(OffsetManagerBase):
     def run(cls, args, cluster_config):
         groups = set()
         kafka_groups = cls.get_kafka_groups(
-            cluster_config, args.use_admin_client,
+            cluster_config,
+            args.use_admin_client,
         )
         if kafka_groups:
             groups.update(kafka_groups)
--- a/kafka_utils/kafka_consumer_manager/commands/list_groups.py
+++ b/kafka_utils/kafka_consumer_manager/commands/list_groups.py
@@ -16,7 +16,7 @@ from __future__ import absolute_import
 from __future__ import print_function
 
 from .offset_manager import OffsetManagerBase
-from kafka_utils.kafka_consumer_manager.util import KafkaGroupReader
+from kafka_utils.kafka_consumer_manager.util import get_kafka_group_reader
 
 
 class ListGroups(OffsetManagerBase):
@@ -31,10 +31,11 @@ class ListGroups(OffsetManagerBase):
         parser_list_groups.set_defaults(command=cls.run)
 
     @classmethod
-    def get_kafka_groups(cls, cluster_config):
+    def get_kafka_groups(cls, cluster_config, use_admin_client=False):
         '''Get the group_id of groups committed into Kafka.'''
-        kafka_group_reader = KafkaGroupReader(cluster_config)
-        return list(kafka_group_reader.read_groups().keys())
+        kafka_group_reader = get_kafka_group_reader(cluster_config, use_admin_client)
+        groups_and_topics = kafka_group_reader.read_groups(list_only=True)
+        return list(groups_and_topics.keys())
 
     @classmethod
     def print_groups(cls, groups, cluster_config):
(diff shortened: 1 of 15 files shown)


flavray:
Example runs:

```
$ time kafka-consumer-manager --cluster-type <> --cluster-name <> offset_get <redacted> --use-admin-client
Cluster name: <>, consumer group: <redacted>
Topic Name: <redacted>  Total Distance: 12067145
	Partition ID: 0
		High Watermark: 12067145
		Low Watermark: 0
		Current Offset: 0
		Offset Distance: 12067145
		Percentage Distance: 100.0%
Topic Name: <redacted>  Total Distance: 1850
	Partition ID: 0
		High Watermark: 197266485
		Low Watermark: 0
		Current Offset: 197264635
		Offset Distance: 1850
		Percentage Distance: 0.0%

real	0m5.371s
user	0m4.232s
sys	0m0.200s


$ time kafka-consumer-manager --cluster-type <> --cluster-name <> offset_get <redacted>
Cluster name: <>, consumer group: <redacted>
Topic Name: <redacted> Total Distance: 421
	Partition ID: 0
		High Watermark: 197269431
		Low Watermark: 0
		Current Offset: 197269010
		Offset Distance: 421
		Percentage Distance: 0.0%

real	4m15.218s
user	3m59.092s
sys	0m11.048s
```

flavray:
Ha. I had forgotten about `OffsetWriter`. The build should be fixed on the next run :)

kafka_utils/kafka_consumer_manager/commands/list_groups.py
```diff
@@ -53,7 +54,9 @@ def print_groups(cls, groups, cluster_config):
     @classmethod
     def run(cls, args, cluster_config):
         groups = set()
-        kafka_groups = cls.get_kafka_groups(cluster_config)
+        kafka_groups = cls.get_kafka_groups(
+            cluster_config, args.use_admin_client,
```
lsterk: Just a note: this doesn't match PEP8 (one arg per line in multi-line functions) but that's fine, since we don't have it as a pre-commit hook...

Maybe some day I'll add it to our pre-commit hooks and get constant style enforcement.

kafka_utils/kafka_consumer_manager/main.py
```diff
@@ -80,6 +80,11 @@ def parse_args():
         action='store_true',
         help='Show DEBUG level logging',
     )
+    parser.add_argument(
+        "--use-admin-client", action="store_true",
+        help="Use the Kafka AdminClient interface to fetch consumer "
+        "group offsets."
```
lsterk: You may consider adding a note about performance, e.g. "Use the Kafka AdminClient to fetch consumer group offsets for quicker performance"

I can imagine that many users of this CLI may not be familiar with the difference, and could benefit from a suggestion that using this is supposed to be significantly faster without any expected downsides (AFAIK, assuming the Kafka broker supports this interface)

tests/kafka_consumer_manager/test_utils.py
```diff
@@ -25,9 +28,88 @@
 Message = namedtuple("Message", ["partition", "offset", "key", "value"])
 
 
+class TestKafkaAdminGroupReader(object):
+
+    @contextlib.contextmanager
```
lsterk: Writing this as a `pytest.fixture` seems preferable, as tests will only have to do one of these options:
```
@pytest.mark.usefixtures("mock_admin_client")
def test(...)
```
or
```
def test_something(self, mock_admin_client):
```
to get the context manager used. This is easier than adding a context manager explicitly in code, and removes one level of indentation!

Since we never use the value yielded by the context manager, it would be better to use the first option.

kafka_utils/kafka_consumer_manager/commands/list_groups.py
flavray: I was following black's guideline, but we've been using PEP8 in this codebase. Good point :)

tests/kafka_consumer_manager/test_utils.py
flavray: Other tests were using contextlib, so I figured it would be good to stay consistent, but heh, let's make the world better by using fixtures :) Thanks!

kafka_utils/kafka_consumer_manager/main.py
flavray: Done :)

PR #256, opened 2020-06-25
Allow dynamic import to find the child-most class

flavray:
When using dynamic import to find a child class of `base_class`, if two
classes match the criteria (e.g: class A, class B(A), class C(B)),
dynamic_import would return B instead of C as it stops at the first
class found in the package that is a subclass of A.

This commit changes the behaviour to return the child-most class.

### DIFF
--- a/kafka_utils/util/utils.py
+++ b/kafka_utils/util/utils.py
@@ -65,11 +65,13 @@ def child_class(class_types, base_class):
         # If more than one class is a subclass of `base_class`
         # It is possible that one or more classes are subclasses of another
         # class (see example above).
-        # Recursively find the child-most class.
+        # Recursively find the child-most class. Break ties by returning any
+        # child-most class.
         for c in subclasses:
             child = child_class(subclasses, c)
             if child is not None:
                 return child
+        return subclasses.pop()
 
 
 def dynamic_import(module_full_name, base_class):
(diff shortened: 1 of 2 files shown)


kafka_utils/util/utils.py
```diff
@@ -34,9 +34,48 @@ def get_module(module_full_name):
...
+    else:
+        # If more than one class is a subclass of `base_class`
+        # It is possible that one or more classes are subclasses of another
+        # class (see example above).
+        # Recursively find the child-most class.
+        for c in subclasses:
+            child = child_class(subclasses, c)
+            if child is not None:
+                return child
+
```
rohangulati, line 73: I am not sure how this function is used but shouldnt we be returning `subclasses.pop()` here in case we are not able to find the most-child subclass recursively. This might happen in the following case
`class A class B(A) class C(A)` where `B` and `C` are not subclasses of each other. I am not sure whether this case is possible, so this is more of a question 😄

flavray: Good point!

flavray: It shouldn't happen (otherwise we'd get a somewhat undefined behaviour), but we "solve" this by making sure only one child-most class exists per module. May as well return at least one result in case that happens.

PR #257, opened 2020-08-17
Fix missing rf_mismatch arg in process_assignment call

lennrt:
## Context
This will allow us to (actually) ignore errors of the form `Mismatch in replication-factor of partitions`. This is particularly useful when issuing commands to recover from incidents related to https://issues.apache.org/jira/browse/KAFKA-8733.

## Changes
- Fix missing rf_mismatch arg in process_assignment
- Fix version conflict in requirements-dev
- Address various `flake8` issues

## Testing Done
`make test` passes


lennrt:
Released in https://github.com/Yelp/kafka-utils/releases/tag/v3.7.1

PR #262, opened 2020-11-06
Change read directory of kafka_discovery files

mhaseebmlk:
This changes the default directory used to read kafka discovery config files from `/etc/kafka_discovery` to `/nail/srv/configs/kafka_discovery_configs`. The actual format of the files does not change.

This will be released with a major version bump.

### Testing
`make test` passes


lennrt (approved):
LGTM

PR #267, opened 2021-01-04
Fix bullet point rendering on kafka_check.rst

andpol:
GitHub and https://kafka-utils.readthedocs.io/en/latest/kafka_check.html doesn't render the bullet points unless there's a newline before the start of the unordered list.


lsterk (approved):
Thanks for working on the docs! 👍

PR #268, opened 2021-01-22
Drop official py27 support

lsterk:
Python2.7 hit End-of-Life in January 2020. Although we have historically tried our best to support Python 2 beyond this date, various other package maintainers have discontinued their support (and explicitly disallow it).

I propose dropping support here (as part of a campaign to fix our open-source Travis build pipeline) but not actively blocking installation into py2. However, we will be unable to test against Python 2 from now on.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-  - 2.7
   - 3.6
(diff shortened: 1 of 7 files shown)


JayH5:
Obsolete

PR #275, opened 2021-06-07
KAFKA-32352: Add replication factor mismatch topics command

hcet:
## Problem
Right now there is no way of finding which topic partitions have number of replicas less than the configured replication factor of the topic.

## Solution
We add a new command to show the user which topics has replication factor mismatch among its partitions.

rf-mismatch

## Verification
tests pass

## Release Plan
`minor` release

### DIFF
--- a/kafka_utils/kafka_cluster_manager/cmds/replication_factor_mismatch_topics.py
+++ b/kafka_utils/kafka_cluster_manager/cmds/replication_factor_mismatch_topics.py
@@ -1,0 +1,52 @@
+# -*- coding: utf-8 -*-
+# Copyright 2021 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from __future__ import absolute_import
+
+from .command import ClusterManagerCmd
+
+
+class ReplicationFactorMismatchTopicsCmd(ClusterManagerCmd):
+
+    def __init__(self):
+        super(ReplicationFactorMismatchTopicsCmd, self).__init__()
+
+    def build_subparser(self, subparsers):
+        subparser = subparsers.add_parser(
+            'replication-factor-mismatch-topics',
+            description='Get topics with replication factor mismatch among its partitions.',
+            help='This command is used to find which topics have partitions with replicas less than the configured '
+                 'replication factor of the topic.',
+        )
+        return subparser
+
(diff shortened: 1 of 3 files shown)


kafka_utils/kafka_cluster_manager/cmds/replication_factor_mismatch_topics.py
```diff
@@ -0,0 +1,52 @@
...
+    def run_command(self, ct, cluster_balancer):
+        rf_mismatch_topics = []
+
+        for topic in ct.topics.values():
+            # Kafka considers first partition's replica count as the replication factor of the topic
+            # So that, any of the partitions having a different replica count means there is a mismatch
+            # https://github.com/apache/kafka/blob/77a89fcf8d7fa01831683327727f11f4bed97319/core/src/main/scala/kafka/admin/TopicCommand.scala#L286
+            partitions_replication_count = {len(partition.replicas) for partition in topic.partitions}
+
+            if len(partitions_replication_count) > 1:
```
zhaoyanh1202, line 49: Sorry I also don't really understand this logic here, is `partitions_replication_count` a dictionary or a list? why do we use length of `partitions_replication_count` > 1 to check whether there is a rf mismatch?

hcet: It's a set. If each partition has the same replication count, then the `len(partitions_replication_count)` should be 1 otherwise it's `>1` and it means there is a mismatch

kafka_utils/kafka_cluster_manager/cmds/replication_factor_mismatch_topics.py
```diff
@@ -0,0 +1,52 @@
...
+            print("\tTopic ID: {topic_id}".format(topic_id=topic.id))
+            print("\t\tPartitions: {partitions}".format(partitions=topic.partitions))
+
+    def run_command(self, ct, cluster_balancer):
+        rf_mismatch_topics = []
+
+        for topic in ct.topics.values():
+            # Kafka considers first partition's replica count as the replication factor of the topic
+            # So that, any of the partitions having a different replica count means there is a mismatch
+            # https://github.com/apache/kafka/blob/77a89fcf8d7fa01831683327727f11f4bed97319/core/src/main/scala/kafka/admin/TopicCommand.scala#L286
```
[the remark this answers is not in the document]
hcet: Topic's replication factor is not stored as a configuration anywhere afaik. If you check our `set_replication_factor` command, we only add or remove replicas from partitions https://github.com/Yelp/kafka-utils/blob/master/kafka_utils/kafka_cluster_manager/cmds/set_replication_factor.py#L108-L111 / https://github.com/Yelp/kafka-utils/blob/master/kafka_utils/kafka_cluster_manager/cmds/set_replication_factor.py#L132-L136

So that, when we get `topic.replication_factor` , it returns the first partition's replica count. Which means there is no difference between checking `len(topic.partitions[0].replicas)` and `topic.replication_factor`

I added this comment and made the code slightly longer to make it clear what we are doing but all my logic can actually fit in 3 lines:
``` python
for topic in ct.topics.values():
    if any((len(partition.replicas) != topic.replication_factor) for partition in topic.partitions):
        rf_mismatch_topics.append(topic)
```

PR #279, opened 2021-11-03
Fix build failing because of dependencies conflict

amreha2002:
Changes:
- Updated `ipdb` and `ipython` as they failed to install because of incompatibility issues with `setuptools==58.3.0`
- Pinned PyYAML in `requirements-dev.txt` as `PyYAML==6.0` is not compatible with `pre-commit==0.8.0` and `docker-compose==1.7.0`

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,3 +1,4 @@
+PyYAML<4,>=3.10 # pre-commit 0.8.0 doesnt support PyYAML>=6.0 and docker-compose 1.7.0 requires PyYAML<4,>=3.10
 cryptography==3.0
 flake8
 importlib-metadata==0.23
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,7 +1,7 @@
 cryptography==3.0
 flake8
 importlib-metadata==0.23
-ipdb==0.10.3
+ipdb==0.13
 ipython==4.2.0
 ipython-genutils==0.1.0
 mock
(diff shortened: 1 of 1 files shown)


requirements-dev.txt
```diff
@@ -1,8 +1,9 @@
+PyYAML<5.1 # pre-commit 0.8.0 doesnt support PyYAML>=6.0 and docker-compose 1.7.0 requires PyYAML<4,>=3.10
```
JayH5, line 1: `PyYAML<4,>=3.10` but we ask for `PyYAML<5.1` and docker-compose isn't in this requirements file?

amreha2002: Changed PyYAML version

amreha2002: docker-compose is defined in tox.ini

JayH5: It'd be nice to upgrade docker-compose and not be stuck with such an old version of PyYAML, but let's move forward for now.

JayH5 (approved):

JayH5:
I'm a bit confused as to why this build passed? https://github.com/Yelp/kafka-utils/pull/273

amreha2002:
@JayH5 
It's probably because we don't install docker-compose or pre-commit except in the `testenv` via tox and requirement-dev.
At this step, PyYAML(6.0) get installed by pre-commit which breaks pre-commit and cause explicit dependency conflict when tox tries installing docker-compose.

PR #281, opened 2021-11-09
When a decommission node is empty, don't crash on max/min functions

lsterk:
If a broker is being decommissioned and it's already empty, we don't have any need to move other topics. Thus, `partitions_to_move` is an empty set.

Currently, the code will crash with an error:
```
[2021-11-09 15:36:05,992] [CRITICAL:main] Uncaught exception:
Traceback (most recent call last):
...
  File "<>/site-packages/kafka_utils/kafka_cluster_manager/cmds/decommission.py", line 112, in run_command
    for partition in partitions_to_move
ValueError: max() arg is an empty sequence
```
I added a unit test to reproduce this issue (it failed without these changes), and it passes now.

My understanding is that if the `partitions_to_move` is empty, then `self.args.max_movement_size ` being 1 doesn't have any impact.

## Release Plan
Patch version release

### DIFF
--- a/kafka_utils/kafka_cluster_manager/cmds/decommission.py
+++ b/kafka_utils/kafka_cluster_manager/cmds/decommission.py
@@ -108,13 +108,11 @@ class DecommissionCmd(ClusterManagerCmd):
             partitions_to_move.update(cluster_topology.brokers[broker].partitions)
 
         largest_size = max(
-            partition.size
-            for partition in partitions_to_move
+            [partition.size for partition in partitions_to_move] or (1, )
         )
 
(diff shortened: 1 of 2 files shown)


lennrt (approved):
Nice fix, LGTM 🚢 

>My understanding is that if the partitions_to_move is empty, then self.args.max_movement_size being 1 doesn't have any impact.

That sounds correct from my perspective as well. After reviewing usage of `max_movement_size` I suppose it would technically still allow partitions of size `0` or `1` to be moved, but there won't be any partitions left to move at that point.

PR #282, opened 2021-11-10
Fix internal build failing

amreha2002:
Changes in requirements-dev:
- Removed `ipdb`, `ipython` dependencies.
- Removed `PyYAML` (it will still get installed by pre-commit(`PyYAML==6.0`), then overwritten when tox installs docker-compose (`PyYAML==3.13`).
- Updated `pre-commit` version to 1.14.4 to make it compatible with`PyYAML==6.0`.


lsterk (approved):
Changes look good to me.

PR #286, opened 2022-10-20
Add type annotations part 1

benbariteau:
This annotates the `kafka_utils.kafka_check` module. It doesn't have any real effect right now, but this repo is large so annotating in chunks makes sense.

I also migrated from retrying to tenacity, since retrying is a dead package and tenacity has type annotations.

### DIFF
--- a/kafka_utils/kafka_check/status_code.py
+++ b/kafka_utils/kafka_check/status_code.py
@@ -11,7 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from __future__ import annotations
+
 import sys
+from typing import Any
+
+from typing_extensions import TypedDict
 
 from kafka_utils.util import print_json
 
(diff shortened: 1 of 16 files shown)


kafka_utils/kafka_check/status_code.py
```diff
@@ -26,14 +31,19 @@
 }
 
 
-def prepare_terminate_message(string):
+class TerminateMessageDict(TypedDict):
+    message: str
+    raw: str
+
+
+def prepare_terminate_message(string: str) -> TerminateMessageDict:
```
JayH5, line 39: It seems like every time we use this function we cast to a `dict[str, Any]` anyway? What's the general approach you are using to decide when to do stricter typing (e.g. `TypedDict` stuff) vs simpler/less code (e.g. just mark the return type as `-> dict` or something)? Or is it all just generated by some tool?

JayH5 (approved):

setup.py
```diff
@@ -58,7 +58,8 @@
         "requests-futures>0.9.0",
         "paramiko>1.8.0,<3.0.0",
         "requests<3.0.0",
-        "retrying",
+        "tenacity",
+        "typing-extensions>=3.7.4",
```
JayH5, line 62: Is the `py.typed` stuff coming in part 2?

benbariteau: Probably a much later part. My guess is 8-10 PRs for typing this.

kafka_utils/kafka_check/status_code.py
benbariteau: That's a good question and I don't have a perfect answer. The reason I used `TerminateMessageDict` in this case is that it has a concrete type. However, the usages of it were more general. Maybe it's not worth it in practice, but I prefer stricter types like this where possible. In theory it makes refactoring this to code to be better designed from a type safety perspective easier.

PR #290, opened 2022-11-01
Add type annotations part 3

benbariteau:
This adds type annotations to all of `kafka_utils.util` and some of `kafka_utils.kafka_consumer_manager.commands`. I started on the latter module before we realized that `kafka_utils.util` was the only module that was meaningfully used.

There's probably more room for better TypedDicts in here (especially in `kafka_utils/util/zookeeper.py`), but I think this is good enough for now.

### DIFF
--- a/kafka_utils/kafka_check/commands/command.py
+++ b/kafka_utils/kafka_check/commands/command.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 import argparse
 from typing import Any
 from typing import cast
+from typing import Dict
 
 from kafka_utils.kafka_check import status_code
 from kafka_utils.kafka_check.status_code import prepare_terminate_message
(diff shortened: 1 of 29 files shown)


amreha2002 (approved):

kafka_utils/kafka_check/commands/command.py
```diff
@@ -58,7 +59,7 @@ def run(self, cluster_config: ClusterConfig, args: argparse.Namespace) -> tuple[
             if args.controller_only and not is_controller(self.zk, args.broker_id):
                 terminate(
                     status_code.OK,
-                    cast(dict[str, Any], prepare_terminate_message(
+                    cast(Dict[str, Any], prepare_terminate_message(
```
JayH5, line 62: What's the reason for switching to `typing.Dict` here?

benbariteau: So `mypy` understands this... but the actual code in python 3.7 doesn't. I ran into an error in the acceptance tests on a cast I added in this branch. And then I realized that I probably broke this code by using this before. So I went back and fixed all `cast` usages.

JayH5 (approved):

